PR #3, opened 2017-07-03
Added initial service and client source code

soccerGB:
Moving the gcs source code from private repo to the newly created opengcs repo


gupta-ak:
So our other repo still has pending PRs, so maybe we wait until later tonight to merge?

soccerGB:
This PR will be replaced by one including the latest changes from the old private repo.

Issue #13, opened 2017-07-07
vsock fds are being accidentally duped to child processes


jstarks:
This is preventing vsock sockets from being closed, which is blocking EOFs from making it back to the host. As a result, docker commit isn't working when in single service VM mode.

This is fixed by linuxkit/virtsock#35. We will need to vendor that in once it is merged.

PR #14, opened 2017-07-07
gcs: Vendor jstarks/virtsock to fix vsock fd leak

jstarks:
This vendors a virtsock fix in so that child processes do not
inherit all vsock fds when they are launched.


jstarks:
Fixes #13. If linuxkit merges the fix soon, I'll update this to vendor back from linuxkit again.

PR #20, opened 2017-07-10
gcs: Improve runtime abstraction

jstarks:
This adds Container and Process interfaces that abstract operations on
container and processes, respectively.

### DIFF
--- a/service/gcs/core/gcs/cleanup.go
+++ b/service/gcs/core/gcs/cleanup.go
@@ -5,21 +5,22 @@ import (
 
 	"github.com/Microsoft/opengcs/service/gcs/oslayer"
 	"github.com/Microsoft/opengcs/service/gcs/prot"
+	"github.com/Microsoft/opengcs/service/gcs/runtime"
 )
 
 // CleanupContainer cleans up the state left behind by the container with the
 // given ID.
 // This function expects containerCacheMutex to be locked on entry.
-func (c *gcsCore) CleanupContainer(id string) error {
+func (c *gcsCore) cleanupContainer(containerEntry *containerCacheEntry) error {
 	var errToReturn error
-	if err := c.forceDeleteContainer(id); err != nil {
+	if err := c.forceDeleteContainer(containerEntry.container); err != nil {
 		logrus.Warn(err)
 		if errToReturn == nil {
 			errToReturn = err
 		}
 	}
 
-	diskMap := c.containerCache[id].MappedVirtualDisks
+	diskMap := containerEntry.MappedVirtualDisks
 	disks := make([]prot.MappedVirtualDisk, 0, len(diskMap))
 	for _, disk := range diskMap {
 		disks = append(disks, disk)
@@ -31,14 +32,14 @@ func (c *gcsCore) CleanupContainer(id string) error {
 		}
 	}
 
-	if err := c.unmountLayers(id); err != nil {
+	if err := c.unmountLayers(containerEntry.ID); err != nil {
 		logrus.Warn(err)
 		if errToReturn == nil {
 			errToReturn = err
 		}
 	}
 
-	if err := c.destroyContainerStorage(id); err != nil {
+	if err := c.destroyContainerStorage(containerEntry.ID); err != nil {
 		logrus.Warn(err)
 		if errToReturn == nil {
 			errToReturn = err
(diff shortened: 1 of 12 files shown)


service/gcs/core/gcs/gcs.go
```diff
@@ -77,6 +77,7 @@ type containerCacheEntry struct {
 	ExitHooks          []func(oslayer.ProcessExitState)
 	MappedVirtualDisks map[uint8]prot.MappedVirtualDisk
 	NetworkAdapters    []prot.NetworkAdapter
+	container          runtime.Container
 }
```
beweedon: You can probably also remove the ID field here since a runtime.Container contains an ID.

service/gcs/core/gcs/gcs.go
```diff
@@ -288,24 +302,28 @@ func (c *gcsCore) ExecProcess(id string, params prot.ProcessParameters, stdioSet
...
 
-	if _, ok := c.containerCache[id]; !ok {
+	containerEntry := c.getContainer(id)
+	if containerEntry == nil {
 		return errors.WithStack(gcserr.NewContainerDoesNotExistError(id))
 	}
 
-	if err := c.Rtime.KillContainer(id, signal); err != nil {
-		return err
+	if containerEntry.container != nil {
```
beweedon: Is there a valid situation where a containerEntry's container field would be nil?

service/gcs/core/gcs/gcs.go
```diff
@@ -359,11 +377,16 @@ func (c *gcsCore) ListProcesses(id string) ([]runtime.ContainerProcessState, err
...
 
-	if _, ok := c.containerCache[id]; !ok {
+	containerEntry := c.getContainer(id)
+	if containerEntry == nil {
 		return nil, errors.WithStack(gcserr.NewContainerDoesNotExistError(id))
 	}
 
-	processes, err := c.Rtime.GetAllContainerProcesses(id)
+	if containerEntry.container == nil {
+		return nil, nil
```
beweedon: Should an error be returned here?

service/gcs/runtime/mockruntime/mockruntime.go
```diff
@@ -13,44 +13,54 @@ import (
 // the container runtime.
 type mockRuntime struct{}
 
+var _ runtime.Runtime = &mockRuntime{}
```
beweedon: More out of curiosity than anything else, but what's the purpose of this?

service/gcs/runtime/runc/runc_test.go
```diff
@@ -34,9 +34,11 @@ func getBundlePath() (string, error) {
 	return filepath.Join(cwd, "testbundle"), nil
 }
 
+var createdContainers []runtime.Container
+
```
beweedon: I'm not sure it's best to have global variables in the tests. The typical pattern used with Ginkgo is to either use local variables in a var block which are included in the close of the It/BeforeEach/etc. functions, or to pass as parameters to helper functions. In this case, since each test should run independently, which its own setup and teardown, it probably makes sense for each test to record its own created containers and pass those into attempKillAndDeleteAllContainers if we don't want to query for all containers on every run.

service/gcs/core/gcs/gcs.go
jstarks: Yes, before the first process is created.

service/gcs/core/gcs/gcs.go
jstarks: I don't think so. When you haven't created the first process, then the list of processes is empty.

service/gcs/runtime/mockruntime/mockruntime.go
jstarks: Ah! This verifies that mockRuntime implements runtime.Runtime so that you get an error here instead of in some test code that instantiates the runtime and tries to pass it as a runtime.Runtime.

service/gcs/runtime/runc/runc_test.go
jstarks: Hmm, I see what you're saying. I was trying to mirror the old behavior (use the file system (global state) to find containers) to something similar (use global variables (global state) to find containers). But I'll see if I can eliminate this altogether.

service/gcs/core/gcs/gcs.go
jstarks: Unfortunately this doesn't work because container is nil until the first process is created.

service/gcs/core/gcs/gcs.go
beweedon: I think it makes more sense to me for a container without any processes in it to still be able to return things like its ID, a list of its processes (which would be empty), etc. That doesn't quite fit with the Linux container model where a container doesn't exist until its first process is created, but it does fit better with the GCS model where a container can be "created" (i.e. its storage and networking set up for example) before a process is created in it. But I'm honestly fine with either approach.

service/gcs/runtime/runc/runc_test.go
beweedon: The reason I had the query to find all existing containers to delete them was because I wanted a stronger guarantee of a clean state after a test finishing running. I didn't want to rely on the tests always reporting back the correct containers, if for instance the code was buggy and a create call reported back the wrong container ID, or actually succeeded in creating a container but returned an error anyway. I know my approach wasn't without its issues anyway (it still needs to call into potentially buggy code to get all existing containers in the first place), but I think it still offers a stronger guarantee of cleaning up everything given buggy create code. I think I'd prefer this method over relying on what containers the create calls claim to create, but I don't feel strongly enough about it to be against your method either.

service/gcs/core/gcs/gcs.go
jstarks: That would be nice, agreed. I'm not sure that the runc package is the best place to make that distinction, though, because runc itself doesn't have the concept of a container without an initial process. We can think about it some more -- my intention with this change is not to change the underlying method flow, just to make it a little more object-oriented. We can change the flow later if we think it will make things cleaner.

beweedon: Sounds good to me!

service/gcs/runtime/runc/runc_test.go
jstarks: I think what I'm about to push is a good compromise.

service/gcs/core/gcs/cleanup.go
```diff
@@ -49,35 +50,35 @@ func (c *gcsCore) CleanupContainer(id string) error {
 }
 
 // forceDeleteContainer deletes the container, no matter its initial state.
-func (c *gcsCore) forceDeleteContainer(id string) error {
-	exists, err := c.Rtime.ContainerExists(id)
+func (c *gcsCore) forceDeleteContainer(container runtime.Container) error {
```
jterry75: Did you mean to pass by ref?

jstarks: `runtime.Container` is an interface backed by `*runc.Container`, so it is by ref.

service/gcs/runtime/runc/runc_test.go
beweedon: Looks good to me!

jstarks:
Repushed with merge conflict resolution.

PR #21, opened 2017-07-10
runc: Close config.json after reading it

jstarks:


jstarks:
This fixes an fd leak.

gupta-ak:
Don't see a close for this one too (line 432)
```go
f, err := os.Create(filepath.Join(tempProcessDir, "process.json"))
```

jstarks:
Good catch

PR #23, opened 2017-07-10
gcs: Pass vsock handles through to external processes

jstarks:
There's no need to use a relay for non-terminal cases; just pass the
vsock handles through. This change only does this for external
processes. A subsequent change will add support in container processes.


service/gcs/core/gcs/gcs.go
```diff
@@ -430,26 +423,36 @@ func (c *gcsCore) RunExternalProcess(params prot.ProcessParameters, stdioSet *co
...
-			cmd.SetStdout(stdioSet.Out)
+		if stdioSet.Out != nil {
+			fileOut, err = stdioSet.Out.File()
+			if err != nil {
+				return -1, errors.Wrap(err, "failed to dup stdout socket for command")
+			}
+			defer fileOut.Close()
+			defer stdioSet.Out.Close()
+			cmd.SetStdout(fileOut)
 		}
```
beweedon: What's the advantage of doing this to just passing stdioSet.Out to cmd.SetStdout?

service/gcs/core/gcs/gcs.go
```diff
@@ -468,35 +471,29 @@ func (c *gcsCore) RunExternalProcess(params prot.ProcessParameters, stdioSet *co
...
 		utils.LogMsgf("external process %d exited with exit status %d", cmd.Process().Pid(), cmd.ExitState().ExitCode())
 
-		// Close stdin so that the copying goroutine is safely unblocked; this is necessary
-		// because the host expects stdin to be closed before it will report process
-		// exit back to the client, and the client expects the process notification before
-		// it will close its side of stdin (which io.Copy is waiting on in the copying goroutine).
-		if stdioSet.In != nil {
-			if err := stdioSet.In.CloseRead(); err != nil {
-				logrus.Errorf("failed call to CloseRead for external process stdin: %v: %s", ociProcess.Args, err)
+		if master != nil {
```
beweedon: The reason before why stdioSet.In was always closed on process exit was because when the process exits, the stdin socket isn't automatically closed. Any copy operations which are hung on a read from that socket never return, because as far as they know more data could still come over the socket. As a result, the relays on the host don't get properly cleaned up. Does this problem not occur when a file is duped and passed in directly? Just want to make sure this is a scenario you tested.

service/gcs/core/gcs/gcs.go
gupta-ak: Since it's an file, you save one fd. The old way would do a ```pipe(2)``` and then send data through the pipe. In this case, it just does a ```dup2```.

service/gcs/core/gcs/gcs.go
jstarks: With the new code, we close stdin after we launch the process (the defers handle this), so there are no copy operations anymore that could block. So when the child process closes stdin, the host gets the notification directly.

service/gcs/core/gcs/gcs.go
jstarks: Exactly.

service/gcs/transport/mock.go
```diff
@@ -46,3 +50,8 @@ func (c *MockConnection) CloseRead() error {
 func (c *MockConnection) CloseWrite() error {
 	return c.PipeWriter.Close()
 }
+
+// File returns a file that can be used to read/write to the connection.
+func (c *MockConnection) File() (*os.File, error) {
+	return nil, errors.New("not supported")
```
beweedon: Nitpicky, but you use the phrase "not implemented" in the error returned from mockReadWriteCloser's File method, but "not supported" here. It probably makes sense to use the same language for both.

jstarks: Done

jstarks:
I resolved the merge.

PR #24, opened 2017-07-10
Support embedding HRESULT codes in errors

beweedon:
This allows the GCS to return HRESULTs to the HCS for certain errors.
HRESULTs are a common way for the HCS to represent errors.

This change only adds HRESULTs to JSON parsing errors at the moment. It
is mostly focused on building the necessary infrastructure. Additional
HRESULTs can be added now as needed.

### DIFF
--- a/service/gcs/bridge/bridge.go
+++ b/service/gcs/bridge/bridge.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/Microsoft/opengcs/service/gcs/core"
+	gcserr "github.com/Microsoft/opengcs/service/gcs/errors"
 	"github.com/Microsoft/opengcs/service/gcs/oslayer"
 	"github.com/Microsoft/opengcs/service/gcs/prot"
 	"github.com/Microsoft/opengcs/service/gcs/transport"
@@ -182,7 +183,7 @@ func (b *bridge) loop() error {
 func (b *bridge) createContainer(message []byte) (*prot.ContainerCreateResponse, error) {
 	response := &prot.ContainerCreateResponse{MessageResponseBase: newResponseBase()}
 	var request prot.ContainerCreate
-	if err := json.Unmarshal(message, &request); err != nil {
+	if err := utils.UnmarshalJSONWithHresult(message, &request); err != nil {
 		return response, errors.Wrapf(err, "failed to unmarshal JSON in message \"%s\"", message)
 	}
 	response.ActivityID = request.ActivityID
@@ -190,7 +191,7 @@ func (b *bridge) createContainer(message []byte) (*prot.ContainerCreateResponse,
 	// The request contains a JSON string field which is equivalent to a
 	// CreateContainerInfo struct.
 	var settings prot.VMHostedContainerSettings
-	if err := json.Unmarshal([]byte(request.ContainerConfig), &settings); err != nil {
+	if err := utils.UnmarshalJSONWithHresult([]byte(request.ContainerConfig), &settings); err != nil {
 		return response, errors.Wrapf(err, "failed to unmarshal JSON for ContainerConfig \"%s\"", request.ContainerConfig)
 	}
 
@@ -215,13 +216,13 @@ func (b *bridge) createContainer(message []byte) (*prot.ContainerCreateResponse,
 func (b *bridge) execProcess(message []byte) (*prot.ContainerExecuteProcessResponse, error) {
 	response := &prot.ContainerExecuteProcessResponse{MessageResponseBase: newResponseBase()}
 	var request prot.ContainerExecuteProcess
-	if err := json.Unmarshal(message, &request); err != nil {
+	if err := utils.UnmarshalJSONWithHresult(message, &request); err != nil {
 		return response, errors.Wrapf(err, "failed to unmarshal JSON for message \"%s\"", message)
 	}
 	// The request contains a JSON string field which is equivalent to an
 	// ExecuteProcessInfo struct.
 	var params prot.ProcessParameters
-	if err := json.Unmarshal([]byte(request.Settings.ProcessParameters), &params); err != nil {
+	if err := utils.UnmarshalJSONWithHresult([]byte(request.Settings.ProcessParameters), &params); err != nil {
 		return response, errors.Wrapf(err, "failed to unmarshal JSON for ProcessParameters \"%s\"", request.Settings.ProcessParameters)
 	}
 
@@ -271,7 +272,7 @@ func (b *bridge) execProcess(message []byte) (*prot.ContainerExecuteProcessRespo
 func (b *bridge) killContainer(message []byte) (*prot.MessageResponseBase, error) {
 	response := newResponseBase()
 	var request prot.MessageBase
-	if err := json.Unmarshal(message, &request); err != nil {
+	if err := utils.UnmarshalJSONWithHresult(message, &request); err != nil {
 		return response, errors.Wrapf(err, "failed to unmarshal JSON for message \"%s\"", message)
 	}
 	response.ActivityID = request.ActivityID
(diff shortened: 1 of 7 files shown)


service/gcs/errors/errors.go
```diff
@@ -1,7 +1,24 @@
...
+	"io"
+
+	"github.com/pkg/errors"
+)
+
+// Hresult is a type corresponding to the HRESULT error type used on Windows.
+type Hresult int32
+
+const (
+	HrUnexpected   = Hresult(-2147418113) // 0x8000FFFF
```
jterry75: Why not just `Hresult(-0x8000FFFF)`?

gupta-ak: ```Hresult(0x8000FFFF)``` won't compile in go due to integer overflow.

beweedon: Yeah. And I need to keep it as a signed value, because otherwise it won't marshal correctly when sent back to the HCS (it will marshal as a positive number rather than a negative one).

jstarks: Hmm. That's unfortunate. Should Hresult be `uint32` instead?

service/gcs/bridge/bridge.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/Microsoft/opengcs/service/gcs/core"
+	gcserr "github.com/Microsoft/opengcs/service/gcs/errors"
```
jstarks: Should open an issue to rename errors to gcserr to avoid the need to do this.

service/gcs/bridge/bridge.go
```diff
@@ -191,6 +193,7 @@ func (b *bridge) createContainer(message []byte) (*prot.ContainerCreateResponse,
 	// CreateContainerInfo struct.
 	var settings prot.VMHostedContainerSettings
 	if err := json.Unmarshal([]byte(request.ContainerConfig), &settings); err != nil {
+		err = gcserr.WrapHresult(err, gcserr.HrVmcomputeInvalidJSON)
```
jstarks: Maybe we need a helper method (unmarshalProtocolJson?) that does an unmarshal and wraps the error with this if it fails.

service/gcs/errors/errors.go
beweedon: If it's unsigned, then the JSON marshaler will marshal it as a positive integer, while the HCS expects it to be negative. I could maybe do some bitwise operations in the bridge to convert between a uint32 Hresult and the int32 the HCS is expecting, that just seemed like more work than was worth.

service/gcs/errors/errors.go
```diff
@@ -45,3 +62,111 @@ func (e *processDoesNotExistError) Error() string {
...
+	hresult Hresult
+}
+
+func (e *baseHresultError) Error() string {
+	return fmt.Sprintf("HRESULT: %d", e.Hresult())
+}
+func (e *baseHresultError) Hresult() Hresult {
+	return e.hresult
+}
+func (e *baseHresultError) Format(s fmt.State, verb rune) {
```
jstarks: Isn't this basically the same as Go's default implementation of Format?

service/gcs/errors/errors.go
```diff
@@ -45,3 +62,111 @@ func (e *processDoesNotExistError) Error() string {
...
+		io.WriteString(s, e.Error())
+	case 'q':
+		fmt.Fprintf(s, "%q", e.Error())
+	}
+}
+func (e *wrappingHresultError) StackTrace() errors.StackTrace {
+	type stackTracer interface {
+		StackTrace() errors.StackTrace
+	}
+	serr, ok := e.Cause().(stackTracer)
```
jstarks: nit: slightly cleaner to invert this:
```
if serr, ok := e.Cause().(stackTracer); ok {
    return serr.StackTrace()
}
return nil

File: service/gcs/bridge/bridge.go (hunk 2 above)
Comment (beweedon) [inline service/gcs/bridge/bridge.go]: Sounds good to me! I can open that issue now!

File: service/gcs/bridge/bridge.go (hunk 3 above)
Comment (beweedon) [inline service/gcs/bridge/bridge.go]: Sure! I can write that!

File: service/gcs/errors/errors.go (hunk 4 above)
Comment (beweedon) [inline service/gcs/errors/errors.go]: Yeah, it might be. I wanted to be explicit, since I defined Format on wrappingHresultError here. For the default formatting behavior to call the Error() method on this, would I need to define a String() method? I could replace this Format method with a String method if that seems correct.

File: service/gcs/errors/errors.go (hunk 5 above)
Comment (beweedon) [inline service/gcs/errors/errors.go]: That's fair. I think I'm going to keep it like this to stick to the general pattern I've seen of returning the requested value at the end of a function, and "invalid" values (such as errors or in this case nil) earlier in the function.

File: service/gcs/errors/errors.go (hunk 4 above)
Comment (jstarks) [inline service/gcs/errors/errors.go]: I think you just need to implement Error(). There's no need to be more explicit since Go guarantees certain behaviors.

https://golang.org/pkg/fmt/ explains the algorithm:

> Except when printed using the verbs %T and %p, special formatting considerations apply for operands that implement certain interfaces. In order of application:
> 
> 1. If the operand is a reflect.Value, the operand is replaced by the concrete value that it holds, and printing continues with the next rule.
> 
> 2. If an operand implements the Formatter interface, it will be invoked. Formatter provides fine control of formatting.
> 
> 3. If the %v verb is used with the # flag (%#v) and the operand implements the GoStringer interface, that will be invoked.
> 
> If the format (which is implicitly %v for Println etc.) is valid for a string (%s %q %v %x %X), the following two rules apply:
> 
> 4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).
> 
> 5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).
>

File: service/gcs/errors/errors.go [hunk 6]
```diff
@@ -45,3 +62,111 @@ func (e *processDoesNotExistError) Error() string {
...
+type StackTracer interface {
+	StackTrace() errors.StackTrace
+}
+
+type baseHresultError struct {
+	hresult Hresult
+}
+
+func (e *baseHresultError) Error() string {
+	return fmt.Sprintf("HRESULT: %d", e.Hresult())
```
jstarks: It would be nice to use %x and cast e.Hresult() to a uint32 so that it looks useful (e.g. 0x8000FFFF instead of -248w49834whatever).

beweedon:
Fixing merge conflicts now.

beweedon:
Merge conflicts fixed. Only conflict was in runc.go, conflict between "defer f.Close()" and changes to JSON encoder line.

service/gcs/core/gcs/storage.go
```diff
@@ -331,7 +330,7 @@ func (c *gcsCore) writeConfigFile(id string, config oci.Spec) error {
 	}
 	defer configFile.Close()
 	writer := bufio.NewWriter(configFile)
-	if err := json.NewEncoder(writer).Encode(config); err != nil {
+	if err := utils.EncodeJSONWithHresult(writer, config); err != nil {
```
jstarks: I'm not sure if we want to use these methods _everywhere_ we deal with JSON. The HRESULT probably should mean something was wrong with the JSON that was passed to GCS from HCS. But in this case, for example, the problem is presumably a logic bug that we can't actually encode an oci.Spec object.

service/gcs/errors/errors.go
```diff
@@ -45,3 +62,101 @@ func (e *processDoesNotExistError) Error() string {
...
+	return e.hresult
+}
+
+type wrappingHresultError struct {
+	cause   error
+	hresult Hresult
+}
+
+func (e *wrappingHresultError) Error() string {
+	return fmt.Sprintf("HRESULT %d", e.Hresult()) + ": " + e.Cause().Error()
```
jstarks: You missed the 0x%x change here.

service/gcs/core/gcs/storage.go
beweedon: That's fair. It makes sense to reserve this HRESULT for errors that are specifically caused by invalid JSON from the host side. I can remove everything that isn't caused by errors parsing host-originating JSON.

beweedon:
Changes have been pushed.

beweedon:
Fixed merge conflicts.

PR #25, opened 2017-07-10
Added a vhdx file with prebuilt empty ext4 for configuing ServiceVM

soccerGB:
Added a vhdx file with prebuilt empty ext4 for configuing ServiceVM


jstarks:
Why does the service VM need a pre-built vhdx?

soccerGB:
The SVM needs to use the disk space as a temporary place for its file conversation operation. It could easily hit out-of-memory for large layers if everything is done on RAM.

jstarks:
OK, but why do we need a .vhdx file in the repo to support this?

soccerGB:
This is to make it easier for the partner to build their own Initrd and vhd to run with our solution.
Ideally, in the future, we could provide instructions on how to build it. For now, to simplify their process, a prebuilt version is provided.

jstarks:
I guess I don't understand the data flow here. Which tool is consuming the vhdx file? Is the vhdx empty when it consumes it, or does the partner add stuff to it?

soccerGB:
This prebuilt sandbox will be consumed by SVM, controlled via DockerD. For each SVM instance, a copy of it will be duplicated for using as the sandbox for a SVM. They starts with empty contents and it will be thrown away after SVM is killed. 
Partners does not need to write anything to it.

jstarks:
Then can't we just have dockerd dynamically create an empty vhdx when it starts up? I would think this would be a small change.

@jhowardmsft?

soccerGB:
Unfortunately, it takes quite sometime (2.8 seconds) to generate such a file, that's why we have it prebuilt.

jhowardmsft:
Right, but in the latest docker bits, it caches the default-size sandbox and scratch files locally under the graphdrivers data-root, so it's just a CopyFileW on the host side. It does not retrieve it every time. The delay to retrieve from an SVM is only when docker:

a) needs to populate the cache or 
b) have a non-default size sandbox. 

However, I though Akash had got this formatting speed -much- lower playing with formatting flags? @gupta-ak

gupta-ak:
@jhowardmsft, the command is 
```
mkfs.ext4 -q -E lazy_itable_init=1 -O ^has_journal,sparse_super2,uninit_bg,^resize_inode <DEVICE>
```

I'm getting like it completed in like 40ms. We can try it to see how long `docker create` takes

gupta-ak:
That only is the mkfs.ext4 time though. There would still be the VHDX hotadd+remove & general docker delay.

jstarks:
I think I'm still missing something. If I understand correctly, we're discussing the difference between two different scenarios:

1. We ship a formatted vhdx file which gets included with the initrd. Docker launches the service VM to extract this thing and caches it locally, then this gets used as the sandbox for containers and for the service VM.

2. Docker dynamically creates and formats a vhdx the first type it creates a service VM. It caches it locally, then this gets used as the sandbox for containers and for the service VM.

The advantage of 2 is that we don't ship a binary blob around that inflates the size of the initrd (which hurts VM start time and VM memory use). I don't know what the advantage of 1 is.

Did I miss anything?

soccerGB:
The advantage of 1 is: it would help reduce the time it takes to boot SVM.

jstarks:
Only the very first time. And it would make all subsequent SVM launches slower because we have to load and decompress the initrd containing the VHDX file.

soccerGB:
Having this prebuiltSandbox in the Initrd is a temporary solution. (it was moved from a separate SVMSandbox file to the initrd file),  ideally it should be stored somewhere on the host along side the OS images.  Once it's moved to the host, it shouldn't have any impact on the Initrd anymore.

jstarks:
Even if we move it out of the initrd, it seems strange to me to have to ship a third artifact. This seems like premature optimization.

The other thing is that I don't believe we will need this for the service VM, just for container sandbox in the future, once the Plan9 functionality is hooked up.

jhowardmsft:
It seems very reasonable (and efficient in terms of initrd size/memory usage) for docker to request/cache it. It's a one-time hit. I can certainly create an empty VHDX, attach it to an SVM and run the format command on it easily enough.

jhowardmsft:
I should add - the current docker logic is to use a second cached VHDX locally for the scratch for the utilities and use the same CopyFileW trick. It just happens it copies the existing pre-existing sandbox to populate the cache.

soccerGB:
For now, we need to get this VHDx  into the repo before partners could build their Linux image to work with our solution.

jstarks:
OK let's put this in until @jhowardmsft can get the dynamic generation work done.

PR #27, opened 2017-07-11
Remove unused vendored packages

jstarks:


soccerGB:
Interesting!
How did we bring those unused ones in? Were they brought in via tools?

jstarks:
I'm not sure. Maybe these used to be important but they aren't anymore.

You can see which ones are unused by running `govendor list +unused`.

PR #31, opened 2017-07-11
Updated patches with previsouly missing Signed-off-by line

soccerGB:
For the following two files:
patch_9pfs_vsock-transport.patch
patch_lower-the-minimum-PMEM-size.patch

### DIFF
--- a/kernelconfig/4.11/patch_9pfs_vsock-transport.patch
+++ b/kernelconfig/4.11/patch_9pfs_vsock-transport.patch
@@ -3,6 +3,8 @@ From: Cheng-mean Liu <soccerl@microsoft.com>
 Date: Tue, 27 Jun 2017 20:04:44 -0700
 Subject: [PATCH] Added vsock transport support to 9pfs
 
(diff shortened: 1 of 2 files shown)


gupta-ak:
Shouldn't the commit hash change once your sign it?

dcui:
I have fixed the hv_sock related commit hash with Cheng-mean in another PR, but I'm not sure if Chean-mean needs to update any hash related to his own files, if any.

gupta-ak:
The top of the patches have a hash:
```
From cbce811777331869e4b2ef3b5cc04de0dc9a0d05 Mon Sep 17 00:00:00 2001
From: Cheng-mean Liu <soccerl@microsoft.com>
```

I think that's the commit hash, so I think these two patches need to be updated with the correct hash. Like this:
```
git commit -s --amend ...
git format-patch ...
```

dcui:
So it looks @soccerGB needs to make a further fix.
Let' wait his reply.

PR #35, opened 2017-07-12
gcs: Pass vsock handles to container processes

jstarks:
This also unifies the TTY relay code for external and container
processes and fixes several possible file descriptor leaks.


jstarks:
@jterry75 is going to finish validating these change before we merge.

service/gcs/bridge/connection.go
```diff
@@ -5,48 +5,18 @@ import (
 
 	"github.com/pkg/errors"
 
+	"io"
```
beweedon: If we're keeping with the patterns of our other imports, "io" should be at the top of the import list, then a newline, then external dependencies, then another newline, then internal dependencies.

service/gcs/bridge/connection.go
```diff
@@ -100,28 +70,30 @@ func sendResponseBytes(conn transport.Connection, messageType prot.MessageIdenti
 // stdio pipe to be used. If CreateStd*Pipe for a given pipe is false, the
 // given Connection is set to nil. After creating each transport.Connection, it
 // calls Connect on it.
-func createAndConnectStdio(tport transport.Transport, params prot.ProcessParameters, settings prot.ExecuteProcessVsockStdioRelaySettings) (*StdioConnSet, error) {
-	var err error
-	var stdin transport.Connection
+func createAndConnectStdio(tport transport.Transport, params prot.ProcessParameters, settings prot.ExecuteProcessVsockStdioRelaySettings) (s *stdio.ConnectionSet, err error) {
```
beweedon: We can probably rename this function to something like "createStdio" now, since it no longer has separate creation and connection stages.

service/gcs/runtime/runc/runc.go
```diff
@@ -376,7 +378,11 @@ func (r *runcRuntime) waitOnProcess(pid int) (oslayer.ProcessExitState, error) {
 }
 
 func (p *process) Wait() (oslayer.ProcessExitState, error) {
-	return p.c.r.waitOnProcess(p.pid)
+	state, err := p.c.r.waitOnProcess(p.pid)
+	if p.relay != nil {
```
beweedon: Since this extra functionality is being added directly to Wait() for *process, maybe Wait() for *container should call process.Wait rather than c.r.waitOnProcess.

service/gcs/stdio/stdio.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	return err
+}
+
+// FileSet contains os.File fields for stdio.
+type FileSet struct {
+	In, Out, Err *os.File
+}
+
+// Close closes all the FIleSet handles.
```
beweedon: Minor, just replace FIleSet with FileSet.

service/gcs/stdio/stdio.go
```diff
@@ -0,0 +1,155 @@
...
+			err = errors.Wrap(cerr, "failed Close on stderr")
+		}
+		fs.Err = nil
+	}
+	return err
+}
+
+// Files returns a FileSet with an os.File for each connection
+// in the connection set.
+func (s *ConnectionSet) Files() (_ *FileSet, err error) {
```
beweedon: I'm not sure how common the pattern used in the return type here is, but it would be clearer at least to me for the return type to be (*FileSet, error), and just have a "var err error" line at the beginning of the function. From what I've read, named return values are mostly intended for documentation rather than functionality.

service/gcs/stdio/stdio.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// Wait waits for the relaying to finish.
+func (r *TtyRelay) Wait() {
+	// Close stdin so that the copying goroutine is safely unblocked; this is necessary
+	// because the host expects stdin to be closed before it will report process
+	// exit back to the client, and the client expects the process notification before
+	// it will close its side of stdin (which io.Copy is waiting on in the copying goroutine).
+	if r.s.In != nil {
+		if err := r.s.In.CloseRead(); err != nil {
```
beweedon: We want this replaced with just plain Close at some point, right? Maybe add a TODO saying that we want to make this change in the future if so.

service/gcs/runtime/runc/runc_test.go
```diff
@@ -182,7 +174,7 @@ var _ = Describe("runC", func() {
 			c  runtime.Container
 		)
 		JustBeforeEach(func() {
-			c, err = rtime.CreateContainer(id, bundle, createAllStdioOptions)
+			c, err = rtime.CreateContainer(id, bundle, &stdio.ConnectionSet{})
```
beweedon: We should probably pass connections into this so that the code which manages writing to them etc. is actually run.

service/gcs/runtime/runc/runc.go
jstarks: Good catch.

service/gcs/stdio/stdio.go
jstarks: No, Close is not safe here because the io.Copy is still outstanding. I think we want to stick with CloseRead unless we switch to using select/poll to make an cancellable io.Copy.

service/gcs/stdio/stdio.go
jstarks: They are also used to ensure that the defer func can evaluate the return value. I'm relying on that here; the defer func needs to only call Close if the err is non-nil.

jstarks: https://blog.golang.org/defer-panic-and-recover

service/gcs/bridge/connection.go
jstarks: I definitely don't like the order that's in this file -- not sure how this happened.

I've been just letting goimports sort the imports when I save the file. By default, goimports will split things into Go runtime dependencies (io, etc.), then non-runtime dependencies (including internal ones), all sorted. This seems to look fine (IMO) and requires less manual intervention than the convention you propose:

```
	"encoding/binary"
	"io"

	"github.com/Microsoft/opengcs/service/gcs/prot"
	"github.com/Microsoft/opengcs/service/gcs/stdio"
	"github.com/Microsoft/opengcs/service/libs/commonutils"
	"github.com/pkg/errors"
```

What do you think?

service/gcs/stdio/stdio.go
beweedon: Oh, yep, I knew that's what it was for. I just thought it would be cleaner if you had a line of code defining err rather than having it implicitly declared in the function signature. I'm fine with it this way, too.

service/gcs/bridge/connection.go
beweedon: Yep, that seems good to me! I didn't know there was a more standard convention, and I'm all for following that rather than the one we've been using in the GCS.

service/gcs/stdio/stdio.go
jstarks: If I declare it in a separate line, then it doesn't work in the case where someone later updates the code to say `return errors.New("whatever")`. If I declare err in a separate line, err will be nil in this case, whereas with named return values, err will be correctly updated to a non-nil value and the func will do the right thing.

service/gcs/bridge/connection.go
jstarks: It looks like goimports will preserve your manual separation if you make one, which is probably why this one got screwed up. But in general I never manually write imports because goimports usually finds the right one for me, so that's why I like their default behavior. If you're using vscode, you can just set a configuration option to use goimports when saving instead of gofmt.

service/gcs/runtime/runc/runc_test.go
jstarks: I see, you mean so that we can exercise this code in the tests, right? It doesn't look like any tests actually validate stdio behavior today, but I might be missing something.

service/gcs/bridge/connection.go
jstarks: What do you think of just "connectStdio"? "create" doesn't seem like the right verb to me.

service/gcs/bridge/connection.go
beweedon: Oh, good advice! I have vim set to use gofmt, but I can change it to use goimports, and we can slowly fix the import formatting in each file as we go.

service/gcs/bridge/connection.go
beweedon: Sounds good to me!

service/gcs/stdio/stdio.go
beweedon: Oh, interesting. That makes sense! Good to know! I thought named return values were just syntactic sugar for declaring on the first line of the function.

service/gcs/runtime/runc/runc_test.go
beweedon: Oh, yeah, there are no tests which explicitly validate it. It would still be good to make sure the code gets exercised, though, so at the very least we'll be able to catch any simple errors.

service/gcs/runtime/runc/runc.go
jstarks: Ah, I missed this -- it already does for its init process but not the other processes. We could update this to keep track of the other processes and wait for them appropriately... But I think it actually doesn't really matter since we don't really care about the relays all finishing before we call back to HCS. We should keep an eye on this -- I'll add a note.

service/gcs/runtime/runc/runc_test.go
jstarks: OK I'll see if I can come up with something.

service/gcs/core/gcs/gcs.go
```diff
@@ -406,82 +383,52 @@ func (c *gcsCore) RunExternalProcess(params prot.ProcessParameters, stdioSet *co
...
-			}
-			defer fileErr.Close()
-			defer stdioSet.Err.Close()
-			cmd.SetStderr(fileErr)
+		fileSet, err := stdioSet.Files()
+		if err != nil {
+			return -1, errors.Wrap(err, "failed to set cmd stdio")
 		}
+		defer fileSet.Close()
+		defer stdioSet.Close()
```
gupta-ak: You probably want a ```err == nil``` guard here. The caller of this function closes stdioSet on failure, so if the cmd.Start() fails, you will close this handle twice. Don't think it will be an issue, but the fix is easy enough:
``` go
defer func() { 
    if err == nil {
         stdioSet.Close()
    }
}() 
```

jstarks: stdio.ConnectionSet is designed to support being closed twice, though.

gupta-ak: Yeah, I see the place where you set the in/out/err to nil. Looks like this is fine then

PR #37, opened 2017-07-12
Merges all logging to use logrus

### LINKED ISSUE #33: Consolidate errors in GCS
We want to remove any of the calls to utils.Log or utils.LogMsg in the GCS (under service/gcs), and replace them with calls to logrus. This makes for a more unified logging experience in the GCS.

We'll also want to change the logrus logging location to the location passed into the GCS as an argument. We can do this by removing ProcessCommandlineOptions and putting the equivalent functionality into a function in the GCS's main.go file. We can then switch to using the value passed as a commandline argument rather than the hard-coded "/tmp/logrus.log".

jterry75:
Resolves #33

### DIFF
--- a/service/gcs/bridge/bridge.go
+++ b/service/gcs/bridge/bridge.go
@@ -90,43 +90,43 @@ func (b *bridge) loop() error {
 		var response interface{}
 		switch header.Type {
 		case prot.ComputeSystemCreateV1:
-			utils.LogMsg("received from HCS: ComputeSystemCreateV1")
+			logrus.Info("received from HCS: ComputeSystemCreateV1")
 			response, err = b.createContainer(message)
 			if err != nil {
 				b.outputError(err)
 			}
 		case prot.ComputeSystemExecuteProcessV1:
-			utils.LogMsg("received from HCS: ComputeSystemExecuteProcessV1")
+			logrus.Info("received from HCS: ComputeSystemExecuteProcessV1")
 			response, err = b.execProcess(message)
 			if err != nil {
 				b.outputError(err)
 			}
 		case prot.ComputeSystemShutdownForcedV1:
-			utils.LogMsg("received from HCS: ComputeSystemShutdownForcedV1")
+			logrus.Info("received from HCS: ComputeSystemShutdownForcedV1")
 			response, err = b.killContainer(message)
 			if err != nil {
 				b.outputError(err)
 			}
 		case prot.ComputeSystemShutdownGracefulV1:
-			utils.LogMsg("received from HCS: ComputeSystemShutdownGracefulV1")
+			logrus.Info("received from HCS: ComputeSystemShutdownGracefulV1")
 			response, err = b.shutdownContainer(message)
 			if err != nil {
 				logrus.Error(err)
 			}
 		case prot.ComputeSystemTerminateProcessV1:
-			utils.LogMsg("received from HCS: ComputeSystemTerminateProcessV1")
+			logrus.Info("received from HCS: ComputeSystemTerminateProcessV1")
 			response, err = b.terminateProcess(message)
 			if err != nil {
 				b.outputError(err)
 			}
 		case prot.ComputeSystemGetPropertiesV1:
-			utils.LogMsg("received from HCS: ComputeSystemGetPropertiesV1")
+			logrus.Info("received from HCS: ComputeSystemGetPropertiesV1")
 			response, err = b.listProcesses(message)
 			if err != nil {
 				b.outputError(err)
 			}
 		case prot.ComputeSystemWaitForProcessV1:
-			utils.LogMsg("received from HCS: ComputeSystemWaitForProcessV1")
+			logrus.Info("received from HCS: ComputeSystemWaitForProcessV1")
 			response, err = b.waitOnProcess(message, header)
 			if err != nil {
 				b.outputError(err)
@@ -136,13 +136,13 @@ func (b *bridge) loop() error {
 				response = nil
 			}
 		case prot.ComputeSystemResizeConsoleV1:
-			utils.LogMsg("received from HCS: ComputeSystemResizeConsoleV1")
+			logrus.Info("received from HCS: ComputeSystemResizeConsoleV1")
 			response, err = b.resizeConsole(message)
 			if err != nil {
 				b.outputError(err)
 			}
 		case prot.ComputeSystemModifySettingsV1:
-			utils.LogMsg("received from HCS: ComputeSystemModifySettingsV1")
+			logrus.Info("received from HCS: ComputeSystemModifySettingsV1")
 			response, err = b.modifySettings(message)
 			if err != nil {
 				b.outputError(err)
@@ -183,7 +183,7 @@ func (b *bridge) loop() error {
 func (b *bridge) createContainer(message []byte) (*prot.ContainerCreateResponse, error) {
 	response := &prot.ContainerCreateResponse{MessageResponseBase: newResponseBase()}
 	var request prot.ContainerCreate
-	if err := utils.UnmarshalJSONWithHresult(message, &request); err != nil {
+	if err := commonutils.UnmarshalJSONWithHresult(message, &request); err != nil {
 		return response, errors.Wrapf(err, "failed to unmarshal JSON in message \"%s\"", message)
 	}
 	response.ActivityID = request.ActivityID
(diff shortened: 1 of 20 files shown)


service/gcs/bridge/bridge.go
```diff
@@ -90,43 +90,43 @@ func (b *bridge) loop() error {
 		var response interface{}
 		switch header.Type {
 		case prot.ComputeSystemCreateV1:
-			utils.LogMsg("received from HCS: ComputeSystemCreateV1")
+			logrus.Print("received from HCS: ComputeSystemCreateV1")
```
jstarks: I'm more used to seeing logrus.Info than logrus.Print (which I think are equivalent but I had to look it up).

service/gcs/bridge/connection.go
```diff
@@ -3,12 +3,13 @@ package bridge
 import (
 	"encoding/binary"
 
+	"github.com/Sirupsen/logrus"
 	"github.com/pkg/errors"
 
+	"io"
```
jstarks: You'll want to manually reorder this to the top. Ben just noticed the same thing in my change :)

service/gcs/main.go
```diff
@@ -15,21 +18,39 @@ import (
...
+		fmt.Fprintf(os.Stderr, "Examples:\n")
+		fmt.Fprintf(os.Stderr, "    %s -loglevel=verbose -logfile=gcslog.txt (default)\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "    %s -loglevel=verbose -logfile=stdout\n", os.Args[0])
+		fmt.Fprintf(os.Stderr, "    %s -loglevel=none \n", os.Args[0])
+	}
+
+	flag.Parse()
+
+	// Use a file instead of stdout
+	if !strings.EqualFold(*logFile, "stdout") {
```
jstarks: It's a little more unixy to make empty string mean stdout instead of a special token like this.

service/gcsutils/gcstools/commoncli/common.go
```diff
@@ -79,6 +80,15 @@ func SetupLogging(args ...*string) error {
 	if len(args) < 1 {
 		return fmt.Errorf("Invalid log params")
 	}
-	utils.SetLoggingOptions("verbose", *args[0])
+
+	logrus.AddHook(commonutils.NewStackHook([]logrus.Level{logrus.DebugLevel}))
```
jstarks: This doesn't seem like quite the right set of cases where we want the hook to fire. It seems like if I turn on debug logging (not sure how that's done in this case), then I want the hook to fire for all levels, not just the debug level.

service/gcs/main.go
jstarks: (Maybe you're just copying existing behavior)

jterry75: No need to maintain existing behavior when its new!

service/gcs/bridge/connection.go
jterry75: Done.

service/gcs/bridge/bridge.go
jterry75: Fixed.

service/gcsutils/gcstools/commoncli/common.go
jterry75: Interesting. I think that we want this then:
```go
if level == logrus.DebugLevel {
    logrus.AddHook(commonutils.NewStackHook(logrus.AllLevels))
}
```

So only if the user says -loglevel:debug do we log stacks but we do it for any log message

service/gcs/main.go
jstarks: I interpreted this to mean you were going to fix it... It looks only half fixed.

service/gcsutils/gcstools/commoncli/common.go
```diff
@@ -79,6 +80,15 @@ func SetupLogging(args ...*string) error {
 	if len(args) < 1 {
 		return fmt.Errorf("Invalid log params")
 	}
-	utils.SetLoggingOptions("verbose", *args[0])
+
+	logrus.AddHook(commonutils.NewStackHook([]logrus.Level{logrus.InfoLevel, logrus.DebugLevel}))
```
jstarks: Why these two levels here?

service/gcs/main.go
jterry75: I have no idea how this happened. I seem to have removed that right after sending the update. Either way it now reads

```go
if *logFile != "" {
```

and the default value from ```flag.String``` is ```""```

service/gcsutils/gcstools/commoncli/common.go
jterry75: It seems that in the old behavior we would only log the stacks for a "verbose" level trace. But for commoncli this is not configurable from the cmdline so we "always" got traces. What this does is maintain that behavior such that any calls to ```Print{f}``` or ```Info{f}``` will get the traces but not any higher levels. 

If you want we can just do ```logrus.AllLevels``` for the hook here but we should probably make the log level configurable then and only enable that if we pass ```DebugLevel``` like I did in ```main.go```

service/gcs/main.go
```diff
@@ -15,21 +17,40 @@ import (
...
+		fmt.Fprintf(os.Stderr, "    %s -loglevel=info -logfile=stdout\n", os.Args[0])
+	}
+
+	flag.Parse()
+
+	// Use a file instead of stdout
+	if *logFile != "stdout" {
+		logFileHandle, err := os.OpenFile(*logFile, os.O_CREATE|os.O_WRONLY, 0600)
+		if err != nil {
+			logrus.Fatalf("%+v", errors.Errorf("failed to create log file %s", *logFile))
```
beweedon: You can probably just use logrus.Fatal and pass in only the error to it here. The %+v is for when an error returned is a github.com/pkg/errors error, where %+v means "print out the error message and its stack trace". Since os.OpenFile doesn't return an error with a stack trace, the %+v isn't necessary.

service/gcs/main.go
```diff
@@ -15,21 +17,40 @@ import (
...
 	}
 
-	// Set logrus output file.
-	// TODO: Consolidate all logs into one file, probably with logrus.
-	const logFilePath = "/tmp/logrus.log"
-	logFile, err := os.OpenFile(logFilePath, os.O_CREATE|os.O_WRONLY, 0600)
+	level, err := logrus.ParseLevel(*logLevel)
 	if err != nil {
-		logrus.Fatalf("%+v", errors.Errorf("failed to create log file %s", logFilePath))
+		logrus.Fatalf("%+v", err)
```
beweedon: Same here about the %+v

service/gcsutils/gcstools/commoncli/common.go
beweedon: Yeah, you're right, it wasn't configurable before :(. I'm not sure your thoughts, John, but I'm cool with not specifying any stack hooks by default, and just having main.go handle that if specified.

jterry75: Ok will remove.

service/gcs/main.go
jterry75: Fixed.

service/gcs/main.go
jterry75: Fixed.

jterry75:
@soccerGB / @gupta-ak - You ok with this?

service/gcs/main.go
```diff
@@ -15,25 +16,44 @@ import (
...
+		logrus.AddHook(commonutils.NewStackHook(logrus.AllLevels))
+	}
+
+	logrus.SetLevel(level)
+
+	logrus.Info("GCS started")
 	tport := &transport.VsockTransport{}
 	rtime, err := runc.NewRuntime()
 	if err != nil {
-		logrus.Fatalf("%+v", err)
```
beweedon: This +v was actually fine. We don't need to get rid of all +vs in formatting. We just don't need them for errors which originate from Go standard libraries. the %+v format specifier is really only useful when errors have a stack trace attached, which errors returned from os.Open and other standard library functions don't have. For errors returned from our own internal GCS code, we should always use %+v.

jterry75: Ah. I see. Will put that back then. Thanks

service/gcs/main.go
```diff
@@ -15,21 +16,40 @@ import (
...
-	if err := utils.ProcessCommandlineOptions(); err != nil {
-		logrus.Fatalf("%+v", err)
+	logLevel := flag.String("loglevel", "warning", "Logging Level: debug, info, warning, error.")
+	logFile := flag.String("logFile", "", "Logging Target: An optional file name/path. Omit for console output.")
+
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "\nUsage of %s:\n", os.Args[0])
+		flag.PrintDefaults()
+		fmt.Fprintf(os.Stderr, "Examples:\n")
+		fmt.Fprintf(os.Stderr, "    %s -loglevel=debug -logfile=gcslog.txt (default)\n", os.Args[0])
```
soccerGB: After this logging switch, what the target log file name? Is the -logfile parameter still relevant?

jterry75: -logfile=<anything> will specify to output the logs to a file. If you omit the parameter you will get logs to stdout. So yes its still relevant unless I misunderstood your question.

soccerGB: There was a change in the recommended logfile name from gcslog.txt to gcs.log in the documentation and the ms Initrd generation process.

Could you change gcslog.txt to gcs.log here?

jterry75: Done.

PR #39, opened 2017-07-13
Add binaries used by docker

jhowardmsft:

@soccerGB Adds the additional binaries not listed which docker needs. Note strictly mkfs.ext4 is shared and already in the list, but I didn't explicitly mention it was required by docker (for the dynamic sandbox management update shortly to be PR'd).

### DIFF
--- a/docs/customosbuildinstructions.md
+++ b/docs/customosbuildinstructions.md
@@ -81,7 +81,7 @@ Here are the expected contents of each subdirectory /file
        /lib/x86_64-linux-gnu/libuuid.so.1
        /lib/modules
 
-7. **/bin** : binaries in this subdir are categorised into three groups
+7. **/bin** : binaries in this subdir are categorised into four groups
         
     - [GCS binaries](gcsbuildinstructions.md/)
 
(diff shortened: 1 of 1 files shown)


docs/customosbuildinstructions.md
```diff
@@ -105,6 +105,12 @@ Here are the expected contents of each subdirectory /file
              /bin/iproute
              /bin/hostname
 
+    - Required binaires: utilities used by docker
+
+             /bin/ls
+             /bin/cat
+             /bin/test
```
soccerGB: I am curious about what /bin/test is for. Is it only for testing purpose?

jhowardmsft: To validate if a directory exists when looking up a disk device from a controller/LUN. Part of the dynamic sandbox management stuff.

soccerGB: I see.
Can we switch to a name not that generic?

jhowardmsft: What do you mean? /bin/test is a standard *nix binary....

soccerGB: I was suggesting instead of "test", use other name dir_validatation_util so something more specific about its purpose.

jhowardmsft: Why? I wouldn't rename `ls` as `list_files_in_some_place_on_disk`, or `cat` as `type_out_contents_of_a_file`....

https://linux.die.net/man/1/test

soccerGB: Oops, I was working a few other file and thinking it's a new binary that we generated. 
Please ignore my previous comments.

PR #44, opened 2017-07-13
GCS log to gcs.log

jhowardmsft:

Consistency..... not `gcslog.txt`, `gcs.log`


gupta-ak:
There's a gcslog.txt in our `.gitignore`. @beweedon Do we need to change it to gcs.log now ?

PR #50, opened 2017-07-14
Fix tests and other minor changes

beweedon:
Tests were failing. These changes should hopefully fix them. Also, moved gcsrunc state from /var/lib to /var/run.

### DIFF
--- a/service/gcs/runtime/runc/runc.go
+++ b/service/gcs/runtime/runc/runc.go
@@ -515,9 +515,15 @@ func (c *container) startProcess(tempProcessDir string, hasTerminal bool, stdioS
 		}
 		defer fileSet.Close()
 		defer stdioSet.Close()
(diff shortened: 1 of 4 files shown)


service/gcs/runtime/runc/runc.go
```diff
@@ -25,7 +25,7 @@ import (
 
 const (
 	runcPath          = "/sbin/runc"
-	containerFilesDir = "/var/lib/gcsrunc"
+	containerFilesDir = "/var/run/gcsrunc"
```
soccerGB: what difference does this change from lib to run make?

beweedon: lib is stored on the disk (usually sda). run however is stored in tmpfs, so it should work even on readonly initrds.

PR #51, opened 2017-07-14
General Cleanup

beweedon:
Just some general cleanup work. Fixed all imports under service/gcs to use the format used by goimports, and removed the now unnecessary printErrors flag from the bridge code.


beweedon:
Merge conflicts resolved

beweedon:
Also, added new commit turning off logging in all test suites, since now that we're using logrus everywhere it will spam output if not turned off.

PR #54, opened 2017-07-15
Fixed a gcs failure due to a logging switching to  logrus package 

soccerGB:
With this bug, the following recommend logging option will fail because "verbose" is not a valid option for logrus logging.

Also updated is the recommended gcs command line in 
scripts\init_script
changed from 
./gcs  -loglevel=verbose -logfile=/tmp/gcs.log & 

to

./gcs  -loglevel=debug -logfile=/tmp/gcs.log &

### DIFF
--- a/kernelconfig/4.11/scripts/init_script
+++ b/kernelconfig/4.11/scripts/init_script
@@ -20,6 +20,6 @@ mount -t tmpfs shm /dev/shm
 
 # Run gcs in the background
 cd /bin
-./gcs  -loglevel=verbose -logfile=/tmp/gcs.log &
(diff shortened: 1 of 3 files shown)


jterry75:
@soccerGB - Why did we set the default level at debug? That seems very 'verbose' for the default. I figured for testing you would start it yourself with the -loglevel:debug so we could leave the in code default to be higher. Thoughts?

soccerGB:
Yes, it is very verbose. That's the recommended for the current state of the GCS, Linux OS partners can easily set to their preferred settings in their init script.  When GCS feature is more stable, we will move the default to less verbose settings.
"verbose" caused break to the logrus.ParseLevel()", which is why "verbose" was replaced here.

PR #57, opened 2017-07-16
gcs: Add per-layer vPMEM support

jstarks:


jstarks:
This change anticipates a host protocol change. This change must be merged before we change the Windows code.

service/gcs/core/gcs/storage.go
```diff
@@ -32,51 +33,87 @@ const (
...
+	layerMounts = make([]*mountSpec, len(layers))
 	for i, layer := range layers {
-		deviceID := layer.Path
-		deviceName, err := c.deviceIDToName(deviceID)
+		deviceName, pmem, err := deviceIDToName(layer.Path)
 		if err != nil {
-			return "", nil, err
+			return nil, nil, err
+		}
+		options := []string{"noload"}
```
jterry75: Should you `const` these strings so we could know the options? Or is this just something that everyone but me knows :)

service/gcs/core/gcs/storage.go
```diff
@@ -32,51 +33,87 @@ const (
...
 
-// deviceIDToName converts from a SCSI location (0, 1, etc.) to a device (sda,
-// sdb, etc.).
-// NOTE: While this function currently works with SCSI codes, it will
-// eventually be reimplemented to use vSCM codes.
-func (c *gcsCore) deviceIDToName(deviceID string) (string, error) {
-	scsiID := fmt.Sprintf("0:0:0:%s", deviceID)
+// scsiLunToName finds the SCSI device with the given LUN. This assumes
+// only one SCSI controller.
+func scsiLunToName(lun int) (string, error) {
```
jterry75: What is the point of changing the type? This is `uint8` on `MappedVirtualDisk` before the cast and makes more sense that way. What does `0:0:0:-1` mean.

service/gcs/core/gcs/storage.go
```diff
@@ -104,32 +141,52 @@ func (c *gcsCore) deviceIDToName(deviceID string) (string, error) {
...
+// deviceIDToName converts a device ID (scsi:<lun> or pmem:<device#> to a
+// device name (/dev/sd? or /dev/pmem?).
+// For temporary compatibility, this also accepts just <lun> for SCSI devices.
+func deviceIDToName(id string) (device string, pmem bool, err error) {
+	const (
+		pmemPrefix = "pmem:"
+		scsiPrefix = "scsi:"
+	)
+
+	if strings.HasPrefix(id, pmemPrefix) {
```
jterry75: Any reason this should be case insensitive?

service/gcs/core/gcs/storage.go
```diff
@@ -104,32 +141,52 @@ func (c *gcsCore) deviceIDToName(deviceID string) (string, error) {
...
 
 // mountMappedVirtualDisks mounts the given disks to the given directories,
 // with the given options. The device names of each disk are given in a
 // parallel slice.
-func (c *gcsCore) mountMappedVirtualDisks(disks []prot.MappedVirtualDisk, devices []string) error {
-	if len(disks) != len(devices) {
-		return errors.Errorf("disk and device slices were of different sizes. disks: %d, devices: %d", len(disks), len(devices))
+func (c *gcsCore) mountMappedVirtualDisks(disks []prot.MappedVirtualDisk, mounts []*mountSpec) error {
+	if len(disks) != len(mounts) {
+		return errors.Errorf("disk and device slices were of different sizes. disks: %d, mounts: %d", len(disks), len(mounts))
```
jterry75: It certainly is the uncommon path so maybe it doesn't matter but given you have already counted once. Just wondering if its better to:

```go
if i, j := len(disks), len(mounts); i != j {
// use I, J here.
}
```

service/gcs/core/gcs/storage.go
```diff
@@ -85,7 +122,7 @@ func (c *gcsCore) deviceIDToName(deviceID string) (string, error) {
 		// Devices matching the given SCSI code should each have a subdirectory
 		// under /sys/bus/scsi/devices/<scsiID>/block.
 		var err error
-		deviceNames, err = c.OS.ReadDir(filepath.Join("/sys", "bus", "scsi", "devices", scsiID, "block"))
+		deviceNames, err = ioutil.ReadDir(filepath.Join("/sys", "bus", "scsi", "devices", scsiID, "block"))
```
beweedon: I agree that we should phase out use of the oslayer interface in the future, and perform mocking for tests at a higher level. However, this change is not the time to do that. This will break the GCS unit tests.

service/gcs/core/gcs/storage.go
```diff
@@ -104,32 +141,52 @@ func (c *gcsCore) deviceIDToName(deviceID string) (string, error) {
 	if len(deviceNames) > 1 {
 		return "", errors.Errorf("more than one block device could match SCSI ID \"%s\"", scsiID)
 	}
-	return deviceNames[0].Name(), nil
+	return "/dev/" + deviceNames[0].Name(), nil
```
beweedon: Use filepath.Join

service/gcs/core/gcs/storage.go
```diff
@@ -32,51 +33,87 @@ const (
...
+	Source     string
+	FileSystem string
+	Flags      uintptr
+	Options    []string
+}
+
+// Mount mounts the file system to the specified target.
+func (ms *mountSpec) Mount(target string) error {
+	options := strings.Join(ms.Options, ",")
+	err := syscall.Mount(ms.Source, target, ms.FileSystem, ms.Flags, options)
```
beweedon: Same here about using the oslayer interface.

service/gcs/core/gcs/storage.go
jstarks: Good point.

service/gcs/core/gcs/storage.go
jstarks: No.

service/gcs/core/gcs/storage.go
jstarks: `len` is an O(1) operation and probably is only invoked once after the Go compiler optimizes, so it's just more typing this way.

I think it's silly that we have this check, but I didn't bother to remove it.

service/gcs/core/gcs/storage.go
jstarks: I don't see where the GCS unit tests are relying on this today.

jstarks:
Please squash this on merge.

soccerGB:
If our partners pick this change without the latest Windows bits, would it break their setup?

beweedon:
Don't merge this yet. Looking at tests.

jstarks:
@soccerGB, no this won't break anything because it still works with the old host protocol.

service/gcs/core/gcs/storage.go
```diff
@@ -31,51 +32,102 @@ const (
...
+	// For now the file system is hard-coded
+	defaultFileSystem = "ext4"
+)
+
+// Mount mounts the file system to the specified target.
+func (ms *mountSpec) Mount(osl oslayer.OS, target string) error {
+	options := strings.Join(ms.Options, ",")
+	err := osl.Mount(ms.Source, target, ms.FileSystem, ms.Flags, options)
+	if err != nil {
+		return errors.Wrapf(err, "mount %s %s %s %x %s", ms.Source, target, ms.FileSystem, ms.Flags, options)
```
beweedon: Maybe stick an "0x" before the %x to make it clear in the output it's hexadecimal

service/gcs/core/gcs/storage.go
```diff
@@ -31,51 +32,102 @@ const (
...
+	mountOptionDax = "dax"
+
+	// For now the file system is hard-coded
+	defaultFileSystem = "ext4"
+)
+
+// Mount mounts the file system to the specified target.
+func (ms *mountSpec) Mount(osl oslayer.OS, target string) error {
+	options := strings.Join(ms.Options, ",")
+	err := osl.Mount(ms.Source, target, ms.FileSystem, ms.Flags, options)
```
beweedon: There's a bug here where ms.Source is only the name of the device to be mounted (so it's sdb or loop0, etc.). The mount system call requires the full path, including "/dev".

service/gcs/core/gcs/storage_test.go
```diff
@@ -286,6 +286,15 @@ var _ = Describe("Storage", func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(string(actualContents)).To(Equal(expectedContents))
 		}
+		scratchSpec := &mountSpec{
+			Source:     "loop0",
+			FileSystem: defaultFileSystem,
+		}
+		layerSpecs := []*mountSpec{
```
beweedon: The scratch space isn't always guaranteed to be assigned to loop0. The same goes for the other layers/loopback devices. Some tests run without a scratch space, some without layers, some without both. scratchSpec and layerSpecs should both be set in the initialization function (BeforeEach), so that the correct specs are used for each test.

service/gcs/core/gcs/storage.go
jstarks: Ah. That's now a test bug -- I want mountSpec to contain the full path to the device, since there's no reason that the source needs to actually start with /dev/.

service/gcs/core/gcs/storage.go
jstarks: OK

jstarks:
Thanks, Ben. I think this latest push should resolve the test issues.

beweedon:
The tests pass now. Looks good to me!

soccerGB:
With this proposed change, will LayersUseVpmem be still needed?
Also, which component (docker, windows) will have control over whether a layer should be represented as vpmem or vhd?

jstarks:
No, LayersUseVpmem is being removed. Docker will be able to control whether an individual layer is vpmem or vhd, but we have no plans on how to expose this right now. The default is vpmem.

soccerGB:
LayersUseVpmem was the way that Docker expresses his preference on pmem over vhd.

With it removed, what's the new replacement for it?
--



--

jstarks:
On a per-layer basis, Docker can set `Cache` to "Enabled" (vPMEM), or the other supported values "Disabled", etc. (SCSI).

This corresponds to vPMEM or SCSI with various caching behaviors. This was designed to support customization of memory partitions. In general, we don't expect the user to care whether we use vPMEM or SCSI.

jstarks:
Can we get this merged? This is blocking host-side work.

soccerGB:
This is a breaking change to my private docker code that uses LayersUseVpmem for instructing the host to expose vPMEM device to the guest.  I need to know what the actual changes that I need to make from Dockerd code to communicate  vPMEM to the host. 
Here is what I used:
		HvRuntime: &hcsshim.HvRuntime{
			ImagePath:       `c:\Program Files\Linux Containers`,
			LinuxKernelFile: `bootx64.efi`,
			LinuxInitrdFile: `initrd.img`,
                        LayersUseVPMEM : true,
...

any example on how to call it after the new changes are in?

jstarks:
With this change and my upcoming host changes, vPMEM will be on by default and works fine. I don't think you need your private Docker code after this...

soccerGB:
That's great to hear.

Issue #52, opened 2017-07-14
Avoid fixed paths to executables like runc


rn:
`gcs` hardcodes the location of things like `runc`. It would be better to use [`exec.LookPath()`](https://golang.org/pkg/os/exec/#LookPath) allowing people to place the binaries in slightly different locations (but still within `PATH)`

rn:
actually, [`exec.Comand`](https://golang.org/pkg/os/exec/#Command) will do `exec.LookPath()` if the command does not contain any path separators. So I renamed the subject of the issue

PR #58, opened 2017-07-17
Removes the fixed paths to runc.

jterry75:
exec.Command will use a LookPath so we dont want to hard code paths to commands we are running.

Resolves: #52


jterry75:
Have run this through the unit tests and they pass

PR #59, opened 2017-07-17
Implement mapped directories

beweedon:
These directories are implemented using the Plan9 filesystem. Given a
mount location and a port, the GCS can mount a directory as a Plan9
share to the host.

### DIFF
--- a/service/gcs/core/gcs/gcs.go
+++ b/service/gcs/core/gcs/gcs.go
@@ -72,6 +72,7 @@ type containerCacheEntry struct {
 	Processes          []int
 	ExitHooks          []func(oslayer.ProcessExitState)
 	MappedVirtualDisks map[uint8]prot.MappedVirtualDisk
+	MappedDirectories  map[uint32]prot.MappedDirectory
 	NetworkAdapters    []prot.NetworkAdapter
 	container          runtime.Container
 }
@@ -80,6 +81,7 @@ func newContainerCacheEntry(id string) *containerCacheEntry {
 	return &containerCacheEntry{
 		ID:                 id,
 		MappedVirtualDisks: make(map[uint8]prot.MappedVirtualDisk),
+		MappedDirectories:  make(map[uint32]prot.MappedDirectory),
 	}
 }
 func (e *containerCacheEntry) AddExitHook(hook func(oslayer.ProcessExitState)) {
@@ -103,6 +105,20 @@ func (e *containerCacheEntry) RemoveMappedVirtualDisk(disk prot.MappedVirtualDis
 		return errors.Errorf("a mapped virtual disk with lun %d is not attached to container %s", disk.Lun, e.ID)
 	}
 	delete(e.MappedVirtualDisks, disk.Lun)
+	return nil
+}
+func (e *containerCacheEntry) AddMappedDirectory(dir prot.MappedDirectory) error {
+	if _, ok := e.MappedDirectories[dir.Port]; ok {
+		return errors.Errorf("a mapped directory with port %d is already attached to container %s", dir.Port, e.ID)
+	}
+	e.MappedDirectories[dir.Port] = dir
+	return nil
+}
+func (e *containerCacheEntry) RemoveMappedDirectory(dir prot.MappedDirectory) error {
+	if _, ok := e.MappedDirectories[dir.Port]; !ok {
+		return errors.Errorf("a mapped directory with port %d is not attached to container %s", dir.Port, e.ID)
+	}
+	delete(e.MappedDirectories, dir.Port)
 	return nil
 }
 
@@ -142,6 +158,10 @@ func (c *gcsCore) CreateContainer(id string, settings prot.VMHostedContainerSett
 	// Set up mapped virtual disks.
 	if err := c.setupMappedVirtualDisks(id, settings.MappedVirtualDisks, containerEntry); err != nil {
 		return errors.Wrapf(err, "failed to set up mapped virtual disks during create for container %s", id)
+	}
+	// Set up mapped directories.
+	if err := c.setupMappedDirectories(id, settings.MappedDirectories, containerEntry); err != nil {
+		return errors.Wrapf(err, "failed to set up mapped directories during create for container %s", id)
 	}
 
 	// Set up layers.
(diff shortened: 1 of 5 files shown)


service/gcs/core/gcs/gcs.go
```diff
@@ -105,6 +107,20 @@ func (e *containerCacheEntry) RemoveMappedVirtualDisk(disk prot.MappedVirtualDis
...
 	return nil
 }
+func (e *containerCacheEntry) AddMappedDirectory(dir prot.MappedDirectory) error {
+	if _, ok := e.MappedDirectories[dir.Port]; ok {
+		return errors.Errorf("a mapped directory with port %d is already attached to container %s", dir.Port, e.ID)
+	}
+	e.MappedDirectories[dir.Port] = dir
+	return nil
+}
+func (e *containerCacheEntry) RemoveMappedDirectory(dir prot.MappedDirectory) error {
```
jterry75: Not sure what we want here but typically I have always in the past had removes be `safe`. IE: Remove something that isn't there is a success because in the end its still the state you want. You could log this if you think its worth debugging but that isnt a failure to me.

service/gcs/core/gcs/gcs.go
```diff
@@ -472,31 +492,39 @@ func (c *gcsCore) ModifySettings(id string, request prot.ResourceModificationReq
 		return errors.WithStack(gcserr.NewContainerDoesNotExistError(id))
 	}
 
+	settings, ok := request.Settings.(prot.ResourceModificationSettings)
```
jterry75: You are actually changing the behavior of the `default` case here but I think its fine. You will just see a different error first.

service/gcs/core/gcs/gcs.go
beweedon: That's fair. That also maintains parity with the unmount code, which only actually performs the unmount if the directory exists and is actually mounted. I can just change this to log an error.

service/gcs/core/gcs/storage.go
```diff
@@ -177,6 +181,51 @@ func (c *gcsCore) unmountMappedVirtualDisks(disks []prot.MappedVirtualDisk) erro
 	return nil
 }
 
+// mountMappedDirectories mounts the given mapped directories using a Plan9
+// filesystem with the given options.
+func (c *gcsCore) mountMappedDirectories(dirs []prot.MappedDirectory) error {
+	for _, dir := range dirs {
```
jterry75: Do we need any cleanup logic here? I guess not really since the container is destroyed yes?

service/gcs/core/gcs/gcs.go
beweedon: Yeah, that's true. I agree, though. At least in this point, as long as the errors are produced, the exact order they appear in is less important.

service/gcs/prot/protocol.go
```diff
@@ -414,6 +421,15 @@ type MappedVirtualDisk struct {
 	ReadOnly          bool  `json:",omitempty"`
 }
 
+// MappedDirectory represents a directory on the host which is mapped to a
+// directory on the guest through a technology such as Plan9.
+type MappedDirectory struct {
+	ContainerPath     string
+	CreateInUtilityVM bool   `json:",omitempty"`
```
jterry75: What is the use of this? Given that we only allow you to pass it as `false` today in `storage.go`

service/gcs/core/gcs/storage.go
beweedon: It's probably a good idea to have cleanup logic there just in case. Same for mountMappedVirtualDisks as well. I didn't include that in this change because adding proper cleanup logic across the GCS seemed like a PR all to itself. Maybe add an issue for that, and we'll tackle it soon?

service/gcs/prot/protocol.go
beweedon: You're right, we could technically leave it out with about the same functionality. I'm keeping it in though because a) we'll probably want to support creating mapped directories and disks inside a container at some point, and b) if we don't produce an error when CreateInUtilityVM is false, then a client on the host side could pass in false, not receive an error, and then go on operating under the assumption the mount they created is actually inside the container, which is behavior we probably don't want.

service/gcs/core/gcs/storage.go
jterry75: Sure: #61

service/gcs/prot/protocol.go
jterry75: Makes sense. Especially `b`

service/gcs/core/gcs/gcs.go
jterry75: Deal.

service/gcs/core/gcs/gcs.go
```diff
@@ -100,12 +100,12 @@ func (e *containerCacheEntry) AddMappedVirtualDisk(disk prot.MappedVirtualDisk)
 	e.MappedVirtualDisks[disk.Lun] = disk
 	return nil
 }
-func (e *containerCacheEntry) RemoveMappedVirtualDisk(disk prot.MappedVirtualDisk) error {
+func (e *containerCacheEntry) RemoveMappedVirtualDisk(disk prot.MappedVirtualDisk) {
 	if _, ok := e.MappedVirtualDisks[disk.Lun]; !ok {
-		return errors.Errorf("a mapped virtual disk with lun %d is not attached to container %s", disk.Lun, e.ID)
+		logrus.Warn("attempt to remove virtual disk with lun %d which is not attached to container %s", disk.Lun, e.ID)
```
jterry75: You want WarnF

service/gcs/core/gcs/gcs.go
```diff
@@ -114,12 +114,12 @@ func (e *containerCacheEntry) AddMappedDirectory(dir prot.MappedDirectory) error
 	e.MappedDirectories[dir.Port] = dir
 	return nil
 }
-func (e *containerCacheEntry) RemoveMappedDirectory(dir prot.MappedDirectory) error {
+func (e *containerCacheEntry) RemoveMappedDirectory(dir prot.MappedDirectory) {
 	if _, ok := e.MappedDirectories[dir.Port]; !ok {
-		return errors.Errorf("a mapped directory with port %d is not attached to container %s", dir.Port, e.ID)
+		logrus.Warn("attempt to remove mapped directory with port %d which is not attached to container %s", dir.Port, e.ID)
```
jterry75: And here too `WarnF`

service/gcs/core/gcs/gcs.go
beweedon: Oh, good catch!

beweedon:
Fixed the Warnf issues

PR #60, opened 2017-07-17
Add script to service/gcs for setting up GCS tests

beweedon:


soccerGB:
Running those unittests requires a matching version of RunC in the system.
I am thinking about vendor-ing the specific version of RunC into the opengcs repo so that we could build/populate RunC as part the unittest test environment. What do you think?

beweedon:
That's probably a good idea, yeah. The unittests could build runc into some temporary directory. To get the runc code to use them, we'd probably have to set the temporary directory as the first path in the $PATH environment variable as well. I can work on that later today or tomorrow hopefully, and put up another PR with the changes.

PR #62, opened 2017-07-17
Added initial Trais CI for the opengcs

soccerGB:
As a starter, only the source code building gate was added. 
Unittest ru will come later


jterry75:
Good start!

PR #63, opened 2017-07-18
Various fixes to the networking setup

### LINKED ISSUE #56: Network config is not transferred into container
I added some additional logging to the networking code 
[0001-Add-logging-to-the-networking-code.patch.txt](https://github.com/Microsoft/opengcs/files/1150315/0001-Add-logging-to-the-networking-code.patch.txt) to debug some networking issue we were seeing.

The log when running `docker.exe run --rm -ti  busybox /bin/sh` in one window and then `hcsdiag.exe read -uvm <id> /tmp/gcs.log` in another window shows after the interface was configured in the root namespace
```
time="2017-07-15T12:20:02Z" level=info msg="/go/src/github.com/Microsoft/opengcs/vendor/github.com/Sirupsen/logrus/entry.go:96 github.com/Microsoft/opengcs/vendor/github.com/Sirupsen/logrus.Entry.log() Configure network adapter 74d16c5a-de09-42f1-ba7a-26b29aed3c13[eth0]: 172.29.14.129/%!s(uint8=20) metric=1 hostip=172.29.0.1, DNSServers=172.29.0.1,10.0.1.1." 
time="2017-07-15T12:20:02Z" level=info msg="/go/src/github.com/Microsoft/opengcs/vendor/github.com/Sirupsen/logrus/entry.go:96 github.com/Microsoft/opengcs/vendor/github.com/Sirupsen/logrus.Entry.log() ifconfig eth0:
eth0      Link encap:Ethernet  HWaddr 00:15:5D:2F:67:C3  
          inet addr:172.29.14.129  Bcast:0.0.0.0  Mask:255.255.240.0
          inet6 addr: fe80::215:5dff:fe2f:67c3/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
```
and before it is moved to the new network namespace:
```
time="2017-07-15T12:20:03Z" level=info msg="/go/src/github.com/Microsoft/opengcs/vendor/github.com/Sirupsen/logrus/entry.go:96 github.com/Microsoft/opengcs/vendor/github.com/Sirupsen/logrus.Entry.log() Move network adapter 74d16c5a-de09-42f1-ba7a-26b29aed3c13[eth0] into namespace 383" 
time="2017-07-15T12:20:03Z" level=info msg="/go/src/github.com/Microsoft/opengcs/vendor/github.com/Sirupsen/logrus/entry.go:96 github.com/Microsoft/opengcs/vendor/github.com/Sirupsen/logrus.Entry.log() ifconfig eth0:
eth0      Link encap:Ethernet  HWaddr 00:15:5D:2F:67:C3  
          inet addr:172.29.14.129  Bcast:0.0.0.0  Mask:255.255.240.0
          inet6 addr: fe80::215:5dff:fe2f:67c3/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:2 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:0 (0.0 B)  TX bytes:176 (176.0 B)
```

So at this point, everything looks ok. However in the container:
```
/ # ifconfig
lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

tunl0     Link encap:UNSPEC  HWaddr 00-00-00-00-30-30-30-30-00-00-00-00-00-00-00-00
          UP RUNNING NOARP  MTU:1480  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
```
The interface is not configured.

Looking at the code in https://github.com/Microsoft/opengcs/blob/master/service/gcs/core/gcs/networking.go#L178, I don't think it is save as Go gives very little control over which OS threads code runs, while the namespace setting is tied to the namespace. See, for example, [this article](https://www.weave.works/blog/linux-namespaces-and-go-don-t-mix) for more details.

I've tried to `runtime.LockOSThread()` at the start of `moveAdapterIntoNamespace()` to no avail. I suspect that by the time that code gets executed there are already a number of threads and go routines running inside the process so this doesn't help.

It's probably safer to start a dedicated helper process to move the interface over to the new namespace. Alternatively, if the network interface is not needed in the root namespace, it would be simpler to just perform the network configuration (ie `configureNetworkAdapter()`) inside the new namespace instead of moving the config.

Full log: 
[gcs.log.txt](https://github.com/Microsoft/opengcs/files/1150318/gcs.log.txt)

rn:
- Factor out networking setup into a separate tool (netnscfg). This is necessary because in Go one has little control over which kernel thread a go routine runs on and one can only enter the current kernel thread into a network namespace. Using a separate utility fixes this.
- Setup the network interfaces directly with the target namespace. This simplifies the network config and makes it more future proof. Previously, the network interface were configured in the root namespace and then the config was move. Setting up the interfaces directly in the target namespace saves us from having to gather the current network config and then trying to recreate it, which would become quite tedious once VXLAN, different MTU sizes etc are supported.
- Handle the case where the network interface may have a different ID inside the namespace.
- Fix resolv.conf. The format and permission were wrong.
- Remove networking and namespace code as it is no longer needed.
- Improve logging of the network setup throughout

I've tested this with LinuxKit and both `nat` and `dhcp` mode of networking config and verified network connectivity from within the container.

resolves #56

### DIFF
--- a/service/gcs/core/gcs/gcs.go
+++ b/service/gcs/core/gcs/gcs.go
@@ -7,6 +7,7 @@ package gcs
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"sync"
 	"syscall"
 	"time"
@@ -153,12 +154,23 @@ func (c *gcsCore) CreateContainer(id string, settings prot.VMHostedContainerSett
 		return errors.Wrapf(err, "failed to mount layers for container %s", id)
 	}
 
-	// Set up networking.
+	// Stash network adapters away
 	for _, adapter := range settings.NetworkAdapters {
-		if err := c.configureNetworkAdapter(adapter); err != nil {
-			return errors.Wrapf(err, "failed to configure network adapter %s", adapter.AdapterInstanceID)
-		}
 		containerEntry.AddNetworkAdapter(adapter)
+	}
+	// Create the directory that will contain the resolv.conf file.
+	//
+	// TODO(rn): This isn't quite right but works. Basically, when
+	// we do the network config in ExecProcess() the overlay for
+	// the rootfs has already been created. When we then write
+	// /etc/resolv.conf to the base layer it won't show up unless
+	// /etc exists when the overlay is created. This is a bit
+	// problematic as we basically later write to a what is
+	// supposed to be read-only layer in the overlay...  Ideally,
+	// dockerd would pass a runc config with a bind mount for
+	// /etc/resolv.conf like it does on unix.
+	if err := c.OS.MkdirAll(filepath.Join(baseFilesPath, "etc"), 0700); err != nil {
+		return errors.Wrapf(err, "failed to create resolv.conf directory")
 	}
 
 	c.containerCache[id] = containerEntry
@@ -194,9 +206,9 @@ func (c *gcsCore) ExecProcess(id string, params prot.ProcessParameters, stdioSet
 		containerEntry.container = container
 		p = container
 
-		// Move the container's network adapters into its namespace.
+		// Configure network adapters in the namespace.
 		for _, adapter := range containerEntry.NetworkAdapters {
-			if err := c.moveAdapterIntoNamespace(container, adapter); err != nil {
+			if err := c.configureAdapterInNamespace(container, adapter); err != nil {
 				return -1, err
 			}
 		}
--- a/service/gcs/core/gcs/networking.go
+++ b/service/gcs/core/gcs/networking.go
@@ -1,18 +1,94 @@
 package gcs
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
-	"syscall"
 
-	"github.com/Microsoft/opengcs/service/gcs/oslayer"
 	"github.com/Microsoft/opengcs/service/gcs/prot"
 	"github.com/Microsoft/opengcs/service/gcs/runtime"
+	"github.com/Sirupsen/logrus"
 	"github.com/pkg/errors"
 )
+
+// configureAdapterInNamespace moves a given adapter into a network
+// namespace and configures it there.
+func (c *gcsCore) configureAdapterInNamespace(container runtime.Container, adapter prot.NetworkAdapter) error {
+	id := adapter.AdapterInstanceID
+	interfaceName, err := c.instanceIDToName(id)
+	if err != nil {
+		return err
+	}
+	nspid := container.Pid()
+	cfg, err := json.Marshal(adapter)
+	if err != nil {
+		return errors.Wrapf(err, "failed to marshal adapter struct to JSON for adapter %s", id)
+	}
+
+	out, err := c.OS.Command("netnscfg",
+		"-if", interfaceName,
+		"-nspid", fmt.Sprintf("%d", nspid),
+		"-cfg", string(cfg)).CombinedOutput()
+	if err != nil {
+		return errors.Wrapf(err, "failed to configure network adapter %s: %s", adapter.AdapterInstanceID, out)
+	}
+	logrus.Debugf("netnscfg output:\n%s", out)
+
+	// Handle resolve.conf
+	// There is no need to create <baseFilesPath>/etc here as it
+	// is created in CreateContainer().
+	resolvPath := filepath.Join(baseFilesPath, "etc/resolv.conf")
+
+	if adapter.NatEnabled {
+		// Set the DNS configuration.
+		if err := c.generateResolvConfFile(resolvPath, adapter); err != nil {
+			return errors.Wrapf(err, "failed to generate resolv.conf file for adapter %s", adapter.AdapterInstanceID)
+		}
+	} else {
+		exists, err := c.OS.PathExists(resolvPath)
+		if err != nil {
+			return errors.Wrapf(err, "failed to check if resolv.conf path already exists for adapter %s", adapter.AdapterInstanceID)
+		}
+		if !exists {
+			if err := c.OS.Link("/etc/resolv.conf", resolvPath); err != nil {
+				return errors.Wrapf(err, "failed to link resolv.conf file for adapter %s", adapter.AdapterInstanceID)
+			}
+		}
+
+	}
+	return nil
+}
+
+// generateResolvConfFile generate a resolve.conf file in $baseFilesPath/etc
+// for the given adapter.
+// TODO: This method of managing DNS will potentially be replaced with another
+// method in the future.
+func (c *gcsCore) generateResolvConfFile(resolvPath string, adapter prot.NetworkAdapter) error {
+	fileContents := ""
+	nameservers := strings.Split(adapter.HostDNSServerList, " ")
+	for i, server := range nameservers {
+		// Limit number of nameservers to 3.
+		if i >= 3 {
+			break
+		}
+		fileContents += fmt.Sprintf("nameserver %s\n", server)
+	}
+	fileContents += fmt.Sprintf("search %s\n", adapter.HostDNSSuffix)
+
+	file, err := c.OS.OpenFile(resolvPath, os.O_CREATE|os.O_WRONLY, 0700)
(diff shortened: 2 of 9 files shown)


service/gcsutils/gcstools/netnsConfig.go
```diff
@@ -0,0 +1,157 @@
...
+	var a prot.NetworkAdapter
+	if err := json.Unmarshal([]byte(*cfgStr), &a); err != nil {
+		return err
+	}
+
+	if a.NatEnabled {
+		log.Infof("Configure %s in %d with: %s/%d gw=%s", *ifStr, *nspid, a.AllocatedIPAddress, a.HostIPPrefixLength, a.HostIPAddress)
+	} else {
+		log.Infof("Configure %s in %s with DHCP", *ifStr, *nspid)
+	}
```
beweedon: Maybe put these log lines down below in the `if a.NatEnabled { ... } else { ... }` block, so they log closer to where the actual work is done.

service/gcsutils/gcstools/netnsConfig.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+		}
+	} else {
+		err := exec.Command(
+			"udhcpc",
+			"-q",
+			"-i", *ifStr,
+			"-s", "/sbin/udhcpc_config.script").Run()
+		if err != nil {
+			return fmt.Errorf("udhcpc failed failed: %v", err)
```
beweedon: Failed is written twice here :)

service/gcs/core/gcs/gcs.go
```diff
@@ -153,13 +154,24 @@ func (c *gcsCore) CreateContainer(id string, settings prot.VMHostedContainerSett
...
+	//
+	// TODO(rn): This isn't quite right but works. Basically, when
+	// we do the network config in ExecProcess() the overlay for
+	// the rootfs has already been created. When we then write
+	// /etc/resolv.conf to the base layer it won't show up unless
+	// /etc exists when the overlay is created. This is a bit
+	// problematic as we basically later write to a what is
+	// supposed to be read-only layer in the overlay...  Ideally,
+	// dockerd would pass a runc config with a bind mount for
+	// /etc/resolv.conf like it does on unix.
```
beweedon: Agreed, using a bind mount would be a good move. One thing we still need to look into RE: DNS is that DNS configuration is passed to the GCS on a per-adapter basis. Since resolv.conf is a per-container, rather than per-adapter, configuration solution, if we want to support per-adapter config we'd maybe have to use something like dnsmasq.

service/gcs/core/gcs/networking.go
```diff
@@ -1,146 +1,63 @@
...
 	if err != nil {
-		return nil, err
+		return err
 	}
-	link, err := c.OS.GetLinkByName(interfaceName)
+	nspid := container.Pid()
+	cfg, err := json.Marshal(adapter)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to get network interface for adapter %s", id)
+		return err
```
beweedon: Use `errors.Wrap` or similar here so we have stack information on the returned error.

service/gcs/core/gcs/networking.go
```diff
@@ -161,85 +78,30 @@ func (c *gcsCore) generateResolvConfFile(adapter prot.NetworkAdapter) error {
...
-		return errors.Wrapf(err, "failed to get the link for adapter %s in preparation for moving it into the container namespace", adapter.AdapterInstanceID)
-	}
-	addrs, err := link.Addrs(syscall.AF_INET)
-	if err != nil {
-		return errors.Wrapf(err, "failed to get the addrs for adapter %s", adapter.AdapterInstanceID)
-	}
-	gateways, err := link.GatewayRoutes(syscall.AF_INET)
+// instanceIDToName converts from the given instance ID (a GUID generated on
+// the Windows host) to its corresponding interface name (e.g. "eth0").
+func (c *gcsCore) instanceIDToName(id string) (string, error) {
```
beweedon: Any reason this was moved to the bottom of the file? Mostly just curious :)

service/gcsutils/gcstools/netnsConfig.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/Microsoft/opengcs/service/gcs/prot"
+	log "github.com/Sirupsen/logrus"
+	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netns"
+)
+
+func netnsConfigMain() {
+	if err := netnsConfig(); err != nil {
+		log.Error("netnsConfig returned: ", err)
```
rn: `logrus.Error()` calls os.Exit(1) according to https://github.com/sirupsen/logrus#level-logging

service/gcsutils/gcstools/netnsConfig.go
rn: I wanted this output at the start, because otherwise, when something errors inbetween this and the other log output I don't know what the config was.

service/gcs/core/gcs/networking.go
rn: Golang seems to have a rough rule of most important functions first. Since I was changing this file anyway, I re-orged the order of the functions

rn:
I've commented on some of the comments and force pushed and update for the other nits. Thanks for the review

service/gcsutils/gcstools/netnsConfig.go
beweedon: It's `logrus.Fatal` which calls `os.Exit`. The comments apply to the lines below them. It also confused me the first time I looked at it, I don't know why. You can also change log.Error to log.Fatal here if you want.

service/gcsutils/gcstools/netnsConfig.go
beweedon: That's fair. I'm cool with this.

service/gcs/core/gcs/networking.go
beweedon: OK, sounds good!

service/gcsutils/gcstools/netnsConfig.go
rn: ah, indeed confusing. Force pushed an update

service/gcs/core/gcs/networking.go
```diff
@@ -1,146 +1,63 @@
...
+		return errors.Wrapf(err, "failed to marshal adapter struct to JSON for adapter %s", id)
 	}
-	return link, nil
-}
 
-// configureNetworkAdapter configures the given network adapter to be used by
-// the container.
-func (c *gcsCore) configureNetworkAdapter(adapter prot.NetworkAdapter) error {
-	link, err := c.getLinkForAdapter(adapter)
+	out, err := c.OS.Command("netnscfg",
```
soccerGB: it looks like this change requiring a new tool netnscfg, any particular version is required?

rn: the tool is included in the previous commit and part of gcsutils

PR #64, opened 2017-07-18
Ci tests

beweedon:
This change gets the GCS tests running on Travis. It includes commits which fix individual problems encountered when trying to run on Travis. The final commit actually runs the tests on Travis. Any suggestions for a particular commit I'll squash back into that commit to keep the PR clean. I'll try to be as explicit as possible in the comments what I changed after feedback, though, so that it's easy to confirm I made the changes requested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,20 @@
 language: go
 
 go:
-- 1.8
+  - 1.8.3
 
 matrix:
   include:
     - os: linux
+      dist: trusty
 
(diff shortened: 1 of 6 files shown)


soccerGB:
It would be cleaner to group the last two commits into its own PR. And have another on for the first few misc changes

beweedon:
Sure, I can do that if you want. I'll close this PR and submit the first 4 as their own PR.

PR #65, opened 2017-07-18
moved temp mount directories away from read-only directory

soccerGB:
This change is to enable the gcsutils set to be able to work on read-only root filesystem

### DIFF
--- a/service/gcsutils/libtar2vhd/tar2vhd.go
+++ b/service/gcsutils/libtar2vhd/tar2vhd.go
@@ -80,7 +80,7 @@ func VHD2Tar(in io.Reader, out io.Writer, options *Options) (int64, error) {
 		return 0, err
 	}
 
-	mntFolder, err := ioutil.TempDir("", "mnt")
+	mntFolder, err := ioutil.TempDir("", "tmp")
 	if err != nil {
 		return 0, err
 	}
--- a/service/gcsutils/tarlib/tardisk.go
+++ b/service/gcsutils/tarlib/tardisk.go
@@ -126,7 +126,7 @@ func CreateTarDisk(in io.Reader,
 
 	logrus.Info("entering CreateTarDisk")
 
-	mntFolder, err := ioutil.TempDir(tmpdir, "mnt")
+	mntFolder, err := ioutil.TempDir(tmpdir, "tmp")
 	if err != nil {
 		return 0, err
 	}


service/gcsutils/libtar2vhd/tar2vhd.go
```diff
@@ -80,7 +80,7 @@ func VHD2Tar(in io.Reader, out io.Writer, options *Options) (int64, error) {
 		return 0, err
 	}
 
-	mntFolder, err := ioutil.TempDir("", "mnt")
```
gupta-ak: the first argument of `ioutil.TempDir` is the directory location, so it should be `ioutil.TempDir("tmp", <whatever>)`

gupta-ak: Also, "" already defaults to "/tmp", so no change is needed here.

service/gcsutils/tarlib/tardisk.go
```diff
@@ -126,7 +126,7 @@ func CreateTarDisk(in io.Reader,
 
 	logrus.Info("entering CreateTarDisk")
 
-	mntFolder, err := ioutil.TempDir(tmpdir, "mnt")
```
gupta-ak: Same here. The real path is `tmpdir`  not `mnt` or `tmp`. Is this causing problems for the read only root file system? `tmpdir` is in a path backed by the scratch disk.

service/gcsutils/libtar2vhd/tar2vhd.go
soccerGB: the 1st parameter is directory and the 2nd parameter is a prefix  to the 1st parameter, which makes the final string no different. Also, I didn't see anywhere that /tmp set as default.

gupta-ak: Think you misread the docs:
```func TempDir(dir, prefix string) (name string, err error)```
TempDir creates a new temporary directory **in the directory dir** with a **name beginning with prefix** and returns the path of the new directory.

See example here: https://play.golang.org/p/bMEYqcs6zN .  The second parameter is for the directory name, not for the where the directory is placed in.

service/gcsutils/tarlib/tardisk.go
soccerGB: I did try running gcstools bits without the changes in this PR and I didn't see the original write error that I was seeing. I might have messed up my environment. I will close this PR for now. And re-open it again with appropriate changes if it's hit again.

PR #66, opened 2017-07-18
Vendored RunC into the opengcs repo to avoid issues from mismatched O…

soccerGB:
…CI spec support

We have seen mismatched  OCI spec support used by gcs, runcs, unitests, and docker daemon.
This vendoring is to eliminate the chance to cause this mismatch among gcs, runc, and unittests.
It also saves the OS image builder the trouble of generating a correct version of runc to run with the opengcs.


soccerGB:
CI caught a build failure,

PR #67, opened 2017-07-18
Refreshes the client code again

jhowardmsft:

This gets the client code up-to-date with the version under jhowardmsft/opengcs which is currently vendored into docker, and with all the dynamic sandbox management stuff working. After I get back from vacation, I'll update docker to point to Microsoft/opengcs and remove the jhowardmsft repo.


jhowardmsft:
@gupta-ak @soccerGB PTAL

PR #70, opened 2017-07-19
Fix permissions on the container root filesystem

### LINKED ISSUE #69: Permission on root filesystem prevent non-root users
Only root has access to `/`:
```
./docker run --rm busybox ls -al /
total 44
drwx------    1 root     root          4096 Jul 19 10:00 .
drwx------    1 root     root          4096 Jul 19 10:00 ..
drwxr-xr-x    2 root     root         12288 Jun 15 00:40 bin
[...]
```
while on Linux:
```
docker run busybox ls -al /
total 44
drwxr-xr-x    1 root     root          4096 Jul 19 10:03 .
drwxr-xr-x    1 root     root          4096 Jul 19 10:03 ..
drwxr-xr-x    2 root     root         12288 Jul 18 19:40 bin
[...]
```
This prevents containers with non-root users to run

rn:
Make sure the scratch and workdir are accessible
by all users.

fixes #69

Signed-off-by: Rolf Neugebauer <rolf.neugebauer@docker.com>

### DIFF
--- a/service/gcs/core/gcs/storage.go
+++ b/service/gcs/core/gcs/storage.go
@@ -216,7 +216,7 @@ func (c *gcsCore) mountLayers(id string, scratchDevice string, layers []string) 
 	if err := c.OS.MkdirAll(baseFilesPath, 0700); err != nil {
 		return errors.Wrapf(err, "failed to create directory for base files %s", baseFilesPath)
 	}
-	if err := c.OS.MkdirAll(scratchPath, 0700); err != nil {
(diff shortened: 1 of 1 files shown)


soccerGB:
Thanks for making this fix. 
We don't have any existing test cases hitting this.  Could you share the scenarios that triggered to hit this bug?

rn:
I tried running `nginx`, which, as many daemons switches users and then couldn't access the default page anymore. But pretty much anything which tries to access something as non-root will fail

PR #72, opened 2017-07-19
Prepped the CI system for running gcs unittests

soccerGB:
Changes include: update Golang from 1.8 to 1.8.3 in in the CI system
Added a run_unittests script for setting up the unttest execution environment (install Runc, creating required directories, and pick up test framework package)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,18 @@
 language: go
 
 go:
-- 1.8
+- 1.8.3
 
 matrix:
   include:
     - os: linux
 
+services:
+    - docker
+
 sudo: required
 
-script: 
-      - cd service && make 
+script:
+     - cd service && make 
+     - cd gcs && ./run_unittests
 
--- a/service/gcs/run_unittests
+++ b/service/gcs/run_unittests
@@ -1,0 +1,31 @@
+#!/bin/bash -ex 
+
+echo "setting up test environment"
+
+# Set the working directory to the script's directory
+script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd -P)
+cd $script_dir
+echo "Generate busybox rootfs directory for runC tests"
+rootfs=runtime/runc/testbundle/rootfs
+mkdir -p $rootfs
+sudo mkdir -p /var/run/gcsrunc
+sudo docker export $(sudo docker create busybox) | tar -C $rootfs -xvf - > /dev/null
+
+echo "clone opencontailers/runc repo and build runc binary"
+
+cd $GOPATH/src/github.com
+mkdir opencontainers && cd opencontainers
+git clone https://github.com/opencontainers/runc runc
+cd runc && git checkout 992a5be178a62e026f4069f443c6164912adbf09
+make BUILDTAGS=``
+sudo cp ./runc /bin/runc
+/bin/runc -version
+
+echo "install Ginkgo"
+go get github.com/onsi/ginkgo/ginkgo
+
+#echo "start running unittest cases"
+#cd $GOPATH/src/github.com/Microsoft/opengcs/service/gcs
+#ginkgo -r -race
+
+


.travis.yml
```diff
@@ -1,14 +1,15 @@
 language: go
 
 go:
-- 1.8
```
beweedon: Do we want to test on both 1.8 and 1.8.3?

service/gcs/run_unittests
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash -ex
```
beweedon: I'm kind of confused. Is this file intended to be a replacement for setup_test_env? We probably want setup and test running to be two separate stages. Running the setup script would go under the `install:` field in the Travis YML file. Then test running would basically just be running ginkgo, and wouldn't even need its own script at this point.

service/gcs/run_unittests
```diff
@@ -0,0 +1,37 @@
...
+mkdir -p $rootfs
+sudo mkdir -p /var/run/gcsrunc
+sudo docker export $(sudo docker create busybox) | tar -C $rootfs -xvf - > /dev/null
+
+echo "clone opencontailers/runc repo and build runc binary"
+
+cd $GOPATH/src/github.com
+mkdir opencontainers && cd opencontainers
+git clone https://github.com/opencontainers/runc runc
+cd runc && git checkout 992a5be178a62e026f4069f443c6164912adbf09
```
beweedon: It probably makes more sense to use `go get -d github.com/opencontainers/runc` instead of `git clone` (the `-d` flag will download runc but not install it). Then, you can cd into the directory, check out the correct commit, and run `go install github.com/opencontainers/runc` to install it into $GOPATH/bin.

service/gcs/run_unittests
```diff
@@ -0,0 +1,37 @@
...
+make BUILDTAGS=``
+sudo cp ./runc /bin/runc
+/bin/runc -version
+
+echo "install Ginkgo and Gomega"
+cd $GOPATH/src/github.com
+go get github.com/onsi/ginkgo/ginkgo
+go get github.com/onsi/gomega
```
beweedon: You don't need to install gomega. Or cd into the $GOPATH/src/github.com directory. Also, this is handled by setup_test_env anyway.

service/gcs/run_unittests
```diff
@@ -0,0 +1,37 @@
...
+
+#echo "start running unittest cases"
+#cd $GOPATH/src/github.com/Microsoft/opengcs/service/gcs
+#ginkgo -r -race
+
+echo "cleaning up test environment"
+cd $GOPATH/src/github.com
+rm -rf opencontainers
+rm -rf onsi
```
beweedon: If we only expect this to run on a Travis VM (which at the moment we do), we don't need cleanup. And, even if this ran on individuals dev machines, we probably wouldn't want to clean up the opencontainers and onsi directories anyway.

service/gcs/run_unittests
soccerGB: It is intended to replace setup_test_env. At this moment, we really only have one set of unittests (gcs) to run. So, it seems be to reasonable to group them into one script.  This is to bring up the CI quality gate as soon as we could to protect the repo from bad checkin.
We will get to all other nice Travis controls later.

service/gcs/run_unittests
soccerGB: I am not sure if we want to put it in the $GOPATH/bin because in the actual guest VM, it's under /bin. I want to make it as close as to the real run environment

service/gcs/run_unittests
soccerGB: I was not sure if the Travis CI will be the only caller of this script. If it is, I could remove the clean up

service/gcs/run_unittests
soccerGB: Thanks for point this out., i will give that a try,

service/gcs/run_unittests
beweedon: Right, yes. I'm fine with running the tests as one script (or one command). But I think it's important to have test setup and test running as two separate stages. For example, if you want to run test setup on your personal dev machine, you should be able to do it without running tests. Travis even has the two stages of setup and testing separated (you've got the `install` field for setup, and the `script` field for testing). I'm not sure I see a reason to stick both pieces of functionality into one script.

service/gcs/run_unittests
beweedon: Well, the point of @jterry75's recent change (https://github.com/Microsoft/opengcs/pull/58) is not to rely on the runc executable being in any particular location, as long as it's in the path. So I don't think it necessarily makes sense to assume runc will be put in /bin or /sbin or wherever.

service/gcs/run_unittests
beweedon: It is at least for now, as far as I know. Maybe if we need it later, we can add a separate cleanup script which Travis doesn't call, but other test environments do.

.travis.yml
soccerGB: only 1.8.3  should be good enough.

service/gcs/run_unittests
soccerGB: I cannot think of a scenario that someone wants to setup the test environment without wanting to run those unittests.

service/gcs/run_unittests
soccerGB: In the real guest environment, /bin is part of the PATH environment variable so it could be launched from any location

.travis.yml
beweedon: OK. For future reference, we should be able to specify each version we want to test on on a separate line if we eventually want to test on multiple versions.

service/gcs/run_unittests
beweedon: I agree. So it shouldn't matter whether runc is placed in $GOPATH/bin or /bin. I really don't have a particularly strong opposition to putting it in /bin, it's just an extra unnecessary step.

service/gcs/run_unittests
beweedon: I mean, they'd want to run them eventually. What I'm saying is that you should be able to set up the test environment without immediately having the tests run. These are two distinct pieces of functionality. And even if we ignore the advantage of having them separate on an individual's dev machine, they are also intentionally separate for Travis. You can see what the difference between the Travis `install` (which installs dependencies, sets up the environment, etc.) and `script` (which runs the build or any tests) are here: https://docs.travis-ci.com/user/customizing-the-build/. Basically, if any step in the `install` phase fails, the entire Travis run exits on an error, because if any stage in setup fails, it's assumed any subsequent tests or builds in the `script` phase will not have the required dependencies to succeed. If any stage in the `script` phase fails, however, the remainder will still run. This allows multiple test suites to be run. We actually already have this now. We have the build portion of `script`, which performs all the building, including linking for gcsutils. We also will have the ginkgo tests. It doesn't make sense that if the building stage fails, the ginkgo tests shouldn't be allowed to run. We should have the build and ginkgo portions of `script` on two separate lines under `script`, and have the setup in the `install` step so that we know that when the `script` phase is reached, the environment has been set up correctly.

`install` isn't so much an advanced Travis option as a standard used across the board in repos which rely on Travis. I hope this makes sense. I'd be happy to talk to you more offline about it!

PR #74, opened 2017-07-20
Test fixes

beweedon:
These changes fix failures in the GCS tests when run on some machines, including the Travis environment. In addition, one change marks the bridge tests as unit tests so they can be run with `ginkgo -focus unittests`.


soccerGB:
Changes look good

Issue #53, opened 2017-07-14
Support AttachOnly flag in MappedVirtualDisk


beweedon:
We only want to actually mount (and unmount) mapped virtual disks if the AttachOnly flag is false. Basic work is currently in the disk_attach_only branch. Still to do is:

- Don't even bother keeping track of disks where AttachOnly is true, so that the if guard in the unmount call is unnecessary
- Add tests
  - Just include the AttachOnly field in bridge_test.go and gcs_test.go
  - Test behavior when AttachOnly is true in storage_test.go

PR #77, opened 2017-07-25
Disk attach only

jterry75:


jterry75:
Resolves #53. Will work on why the storage tests are failing after this to unblock.

service/gcs/prot/protocol.go
```diff
@@ -306,14 +306,6 @@ func UnmarshalContainerModifySettings(b []byte) (*ContainerModifySettings, error
...
 	}
 
-	// Fill in default fields.
-	if request.Request.ResourceType == "" {
-		request.Request.ResourceType = PtMemory
-	}
-	if request.Request.RequestType == "" {
-		request.Request.RequestType = RtAdd
-	}
-
```
jterry75: For `RequestType` I guess you are correct. We should keep the removal of `ResourceType` though because its not `[OmitEmpty]`

beweedon: Oh, true! Sounds good to me!

service/gcs/bridge/bridge.go
```diff
@@ -372,11 +372,20 @@ func (b *bridge) resizeConsole(message []byte) (*prot.MessageResponseBase, error
 
 func (b *bridge) modifySettings(message []byte) (*prot.MessageResponseBase, error) {
 	response := newResponseBase()
+
+	var base prot.MessageBase
```
soccerGB: Is it only for the purpose of getting the ActivityID out?
Is possible to collapse both Unmashall* into one?

jterry75:
I thought thats what the CI was for?

beweedon:
Cheng-mean means running a docker scenario end-to-end to test out the changes. @soccerGB, The commit I added to this PR has already been tested by John Howard, and it works. The only other non-testing change is the activity ID work, which passes our tests and shouldn't cause a problem. If we want to verify again in an end-to-end scenario for this change, we can do that. But also if we want this change in sooner rather than later just wanted you to know end-to-end testing has already been done for the crucial parts of it :)

soccerGB:
Unfortunately, our current CI only  has buddy build and unittest runs validation. 
Before we have an end-2-end scenario tests in place, we still need to run some sanity check on the changes for the end-to-end scenario (at least some very basic one such pull and run busybox ).

soccerGB:
Thanks a lot for the validation

Issue #42, opened 2017-07-13
initrd size grew between 7/3 and ~7/7 by 1.5MB


jhowardmsft:
As sent via email to Cheng-mean, Ben, Akash & John, but no-one replied. Hence logging it here as an issue.

On 7/7 I noticed that it had increased by ~1.5MB. Currently (7/13) it is 9.5M. On 7/3 (master build 411) it was 'only' 8MB.

gupta-ak:
It's a leftover remotefs binary. Should be fixed when remotefs is in master

gupta-ak:
Now it's ~8MB as expected.

PR #81, opened 2017-07-28
Added remotefs binary

gupta-ak:
Implementation of the remote file system for basic `docker cp` and `docker build` functionality.


gupta-ak:
FYI this should also resolve #42.

service/gcsutils/gcstools/remotefs_main.go
```diff
@@ -68,6 +71,21 @@ func getTarOpts(r io.Reader) (*archive.TarOptions, error) {
...
+	if err := binary.Write(b, binary.BigEndian, size); err != nil {
+		return nil, err
+	}
+
+	if _, err := b.Write(rawJSON); err != nil {
+		return nil, err
+	}
+
+	if err := ioutil.WriteFile("/tmp/test.data", b.Bytes(), 0644); err != nil {
+		return nil, err
```
soccerGB: Is this for testing only?

service/gcsutils/gcstools/remotefs_main.go
```diff
@@ -186,6 +187,18 @@ func archivePath() error {
 	return nil
 }
 
+func mkdirall() error {
+	if len(os.Args) < 4 {
+		return InvalidParams
+	}
+
+	perm, err := strconv.ParseUint(os.Args[3], 8, 32)
```
soccerGB: are those dirs being made all on the sandbox?

service/gcsutils/remotefs/remotefs_os.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"encoding/json"
+	"io"
+	"os"
+	"strconv"
+)
+
+// TODO: @gupta-ak. Continue adding as needed to support more docker functionality.
+
```
soccerGB: Are there any more commands to support other then the following already implemented ones? 
"stat" "lstat" "mkdirall" "resolvepath" "extractarchive" "archivepath
If there are, we should list them here, otherwise we can remove this comment.

service/gcsutils/remotefs/utils.go
```diff
@@ -0,0 +1,151 @@
...
+	b, err := ioutil.ReadAll(in)
+	if err != nil {
+		return err
+	}
+
+	var exportedErr ExportedError
+	if err := json.Unmarshal(b, &exportedErr); err != nil {
+		return err
+	}
+
```
soccerGB: It might not be a bad idea to log some error if it gets to the switch case

service/gcsutils/remotefs/types.go
```diff
@@ -22,7 +22,7 @@ type FileInfo struct {
 	NameVar    string
 	SizeVar    int64
 	ModeVar    os.FileMode
-	ModTimeVar time.Time
+	ModTimeVar int64 // Serialization of time.Time breaks in travis, so use an int
```
soccerGB: compilation error? only on Travis CI?

service/gcsutils/remotefs/remotefs_os.go
```diff
@@ -0,0 +1,74 @@
...
```
soccerGB: Are there any more commands to support other then the following already implemented ones?
"stat" "lstat" "mkdirall" "resolvepath" "extractarchive" "archivepath
If there are, we should list them here, otherwise we can remove this comment.

service/gcsutils/remotefs/types.go
gupta-ak: Yeah, it breaks in travis, but not on my local machine. It has something to do with the locale/location. not sure what

service/gcsutils/remotefs/remotefs_os.go
gupta-ak: I'll add them as constants to `service/gcsutils/remotefs/remotefs.go`

service/gcsutils/remotefs/types.go
soccerGB: Could it be because the go version ? The Go version on the Travis is 1.8.3. What's your version?

gupta-ak: same version (1.8.3)

gupta-ak: oh, btw, it's not a compilation error. It's serialization/deserialization issue that causes my unit test to fail on travis but not my local machine

PR #82, opened 2017-07-28
Change TerminateProcess to SignalProcess

beweedon:
Now the GCS will receive a SignalProcessOptions field in the
ContainerSignalProcess message (which was renamed from
ContainerTerminateProcess). Rather than sending a SIGTERM to the process
followed by a SIGKILL after a timeout, the GCS will simply send the
specified signal to the process. Renaming has also been done throughout
the code to match this semantic change.

### DIFF
--- a/service/gcs/bridge/bridge.go
+++ b/service/gcs/bridge/bridge.go
@@ -101,9 +101,9 @@ func (b *bridge) loop() error {
 			if err != nil {
 				logrus.Error(err)
 			}
-		case prot.ComputeSystemTerminateProcessV1:
-			logrus.Info("received from HCS: ComputeSystemTerminateProcessV1")
-			response, err = b.terminateProcess(message)
+		case prot.ComputeSystemSignalProcessV1:
+			logrus.Info("received from HCS: ComputeSystemSignalProcessV1")
+			response, err = b.signalProcess(message)
 			if err != nil {
 				logrus.Error(err)
 			}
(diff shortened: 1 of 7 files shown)


beweedon:
A side effect of this change is that the slow unit tests go away, if that was something that was annoying you :)

service/gcs/bridge/bridge.go
```diff
@@ -268,15 +268,15 @@ func (b *bridge) shutdownContainer(message []byte) (*prot.MessageResponseBase, e
...
 	response := newResponseBase()
-	var request prot.ContainerTerminateProcess
+	var request prot.ContainerSignalProcess
 	if err := commonutils.UnmarshalJSONWithHresult(message, &request); err != nil {
 		return response, errors.Wrapf(err, "failed to unmarshal JSON for message \"%s\"", message)
 	}
 	response.ActivityID = request.ActivityID
 
-	if err := b.coreint.TerminateProcess(int(request.ProcessID)); err != nil {
+	if err := b.coreint.SignalProcess(int(request.ProcessID), request.Options); err != nil {
```
jstarks: We should handle the case where signal is missing since older hosts won't have the signal set. There's no reason for that host change to be a breaking change.

beweedon: That's fair. My reasoning for not doing this is Go doesn't entirely support optional JSON fields, and the best way (at least the best way I know, there might be better ways) of handling this would be to use json.RawMessage as the field type for Signal in order to detect if it's present or not. Either that or special case the default 0 Signal value to mean kill. Both of these seemed hacky enough that it wasn't worth putting them into the GCS to support older RS3 Windows versions. However, if we don't want a breaking change, I can put those changes into the GCS, maybe temporarily even.

jstarks: You can either special case 0 or you can use `*int32` as the type in the struct.

I would just special case 0 like Docker does.

beweedon: Sounds good. I'll add a TODO on it to potentially remove the special casing in the future.

PR #83, opened 2017-07-28
Mount MappedDirectories under full path ContainerPath

beweedon:
Previously, ContainerPath specified a single path element and was always
mounted under /tmp. Now, it is a full path just like
MappedVirtualDisks.

### DIFF
--- a/service/gcs/core/gcs/storage.go
+++ b/service/gcs/core/gcs/storage.go
@@ -23,10 +23,6 @@ const (
 	// that will be used as the base layer for containers.
 	baseFilesPath = "/tmp/base/"
 
-	// mappedDirectoryPrefix is the directory under which all mapped
-	// directories are mounted.
-	mappedDirectoryPrefix = "/tmp"
-
 	// deviceLookupTimeout is the amount of time before deviceIDToName will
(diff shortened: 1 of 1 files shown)


service/gcs/core/gcs/storage.go
```diff
@@ -266,7 +262,7 @@ func (c *gcsCore) mountMappedDirectories(dirs []prot.MappedDirectory) error {
 		if !dir.CreateInUtilityVM {
 			return errors.New("we do not currently support mapping directories inside the container namespace")
 		}
-		mountedPath := filepath.Join(mappedDirectoryPrefix, dir.ContainerPath)
+		mountedPath := dir.ContainerPath
```
jterry75: just use dir.ContainerPath in the calls. Why do the copy?

beweedon: This was honestly just because `mountedPath` is more concise than `dir.ContainerPath`, and strings in Go are just pointers and lengths, so the copy should be pretty inexpensive. If we want, I can just use dir.ContainerPath/disk.ContainerPath directly in storage.go, though. I don't particularly care either way.

PR #84, opened 2017-07-31
Adds console resize support.

jterry75:
Resolves: #76

### DIFF
--- a/service/gcs/bridge/bridge_test.go
+++ b/service/gcs/bridge/bridge_test.go
@@ -587,7 +587,7 @@ var _ = Describe("Bridge", func() {
 		Describe("calling resizeConsole", func() {
 			var (
 				response prot.MessageResponseBase
-				//callArgs mockcore.ResizeConsoleCall
+				callArgs mockcore.ResizeConsoleCall
 			)
 			BeforeEach(func() {
 				messageType = prot.ComputeSystemResizeConsoleV1
@@ -597,7 +597,7 @@ var _ = Describe("Bridge", func() {
 				err := json.Unmarshal([]byte(responseString), &response)
 				Expect(err).NotTo(HaveOccurred())
 				responseBase = &response
-				//callArgs = coreint.LastResizeConsole
+				callArgs = coreint.LastResizeConsole
 			})
 			Context("the message is normal ASCII", func() {
 				BeforeEach(func() {
@@ -613,10 +613,31 @@ var _ = Describe("Bridge", func() {
 				})
 				AssertNoResponseErrors()
 				AssertActivityIDCorrect()
-				// TODO: Add tests on callArgs when resizing the console is
-				// implemented.
 				It("should receive the correct values", func() {
-					//e.g. Expect(callArgs.Pid).To(Equal(101))
+					Expect(callArgs.ID).To(Equal(containerID))
+					Expect(callArgs.Pid).To(Equal(101))
+					Expect(callArgs.Height).To(Equal(uint16(30)))
+					Expect(callArgs.Width).To(Equal(uint16(72)))
+				})
+			})
+			Context("the message is normal ASCII no containerid", func() {
+				BeforeEach(func() {
+					message = prot.ContainerResizeConsole{
+						MessageBase: &prot.MessageBase{
+							ActivityID: activityID,
+						},
+						ProcessID: 102,
+						Height:    80,
+						Width:     80,
+					}
+				})
+				AssertNoResponseErrors()
+				AssertActivityIDCorrect()
+				It("should receive the correct values", func() {
+					Expect(callArgs.ID).To(Equal(""))
+					Expect(callArgs.Pid).To(Equal(102))
+					Expect(callArgs.Height).To(Equal(uint16(80)))
+					Expect(callArgs.Width).To(Equal(uint16(80)))
 				})
 			})
 		})
--- a/service/gcs/bridge/bridge_test.go
+++ b/service/gcs/bridge/bridge_test.go
@@ -614,30 +614,9 @@ var _ = Describe("Bridge", func() {
 				AssertNoResponseErrors()
 				AssertActivityIDCorrect()
 				It("should receive the correct values", func() {
-					Expect(callArgs.ID).To(Equal(containerID))
 					Expect(callArgs.Pid).To(Equal(101))
 					Expect(callArgs.Height).To(Equal(uint16(30)))
 					Expect(callArgs.Width).To(Equal(uint16(72)))
-				})
-			})
-			Context("the message is normal ASCII no containerid", func() {
-				BeforeEach(func() {
-					message = prot.ContainerResizeConsole{
-						MessageBase: &prot.MessageBase{
-							ActivityID: activityID,
-						},
-						ProcessID: 102,
-						Height:    80,
-						Width:     80,
-					}
-				})
-				AssertNoResponseErrors()
-				AssertActivityIDCorrect()
-				It("should receive the correct values", func() {
-					Expect(callArgs.ID).To(Equal(""))
-					Expect(callArgs.Pid).To(Equal(102))
-					Expect(callArgs.Height).To(Equal(uint16(80)))
-					Expect(callArgs.Width).To(Equal(uint16(80)))
 				})
 			})
 		})
--- a/service/gcs/core/gcs/gcs.go
+++ b/service/gcs/core/gcs/gcs.go
@@ -118,7 +118,7 @@ func (e *containerCacheEntry) RemoveMappedDirectory(dir prot.MappedDirectory) {
 type processCacheEntry struct {
 	ExitStatus oslayer.ProcessExitState
 	ExitHooks  []func(oslayer.ProcessExitState)
-	Master     *os.File
+	Tty        *stdio.TtyRelay
 }
 
 func newProcessCacheEntry() *processCacheEntry {
@@ -219,7 +219,7 @@ func (c *gcsCore) ExecProcess(id string, params prot.ProcessParameters, stdioSet
 
 		containerEntry.container = container
 		p = container
-		processEntry.Master = p.Console()
+		processEntry.Tty = p.Tty()
 
 		// Configure network adapters in the namespace.
 		for _, adapter := range containerEntry.NetworkAdapters {
@@ -267,7 +267,7 @@ func (c *gcsCore) ExecProcess(id string, params prot.ProcessParameters, stdioSet
 		if err != nil {
 			return -1, err
 		}
-		processEntry.Master = p.Console()
+		processEntry.Tty = p.Tty()
 
 		go func() {
 			state, err := p.Wait()
@@ -403,10 +403,10 @@ func (c *gcsCore) RunExternalProcess(params prot.ProcessParameters, stdioSet *st
 	cmd.SetEnv(ociProcess.Env)
 
 	var relay *stdio.TtyRelay
-	var master *os.File
 	if params.EmulateConsole {
 		// Allocate a console for the process.
 		var (
+			master      *os.File
 			consolePath string
 		)
 		master, consolePath, err = stdio.NewConsole()
@@ -449,7 +449,7 @@ func (c *gcsCore) RunExternalProcess(params prot.ProcessParameters, stdioSet *st
 	}
 
 	processEntry := newProcessCacheEntry()
-	processEntry.Master = master
+	processEntry.Tty = relay
 	go func() {
 		if err := cmd.Wait(); err != nil {
 			// TODO: When cmd is a shell, and last command in the shell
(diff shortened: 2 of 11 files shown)


jterry75:
I have verified this through docker for both the init process and an exec process such as VI that the console resize is indeed happening.

jstarks:
What happens if the TTY relay finishes and closes master just as the resize console call is coming down?

It seems like some synchronization might be missing there.

jterry75:
I certainly could add synchronization but I didn't for two reasons.
1. If the **resize** is right before a `close` it will succeed. If the **resize** is right after a `close` you will get an error from the `ioctl` call which is no different than returning an error "invalid process" or whatever from locking and detecting the case. IE: the kernel is already a form of synchronization.
2. What is the case where you have a TTY exit the process and then resize it? The TTY case as far as I know is only for interactive sessions in which I am not physically fast enough to exit the process while resizing it.

All that said I am happy to add a mutex on the relay that exposes the console handle. Let me know if you still want it.

jstarks:
The kernel is not a safe form of synchronization. The fd can get reused right after the close call, so you will send the ioctl to the wrong place.

Why can't the relay implement the resize method? Then you can synchronize internally without exposing a mutex.

service/gcs/core/gcs/gcs.go
```diff
@@ -553,18 +562,28 @@ func (c *gcsCore) RegisterContainerExitHook(id string, exitHook func(oslayer.Pro
...
+		ok    bool
+	)
 
-	var entry *processCacheEntry
-	var ok bool
-	entry, ok = c.processCache[pid]
-	if !ok {
-		entry, ok = c.externalProcessCache[pid]
-		if !ok {
+	if id == "" {
```
beweedon: Do we actually expect the HCS to send an empty containerID string for external processes? I didn't know about this behavior, just making sure this is something we can definitely rely on.

service/gcs/core/mockcore/mockcore.go
```diff
@@ -106,7 +115,7 @@ func (c *MockCore) SignalContainer(id string, signal oslayer.Signal) error {
 }
 
 // SignalProcess captures its arguments and returns a nil error.
-func (c *MockCore) SignalProcess(pid int, options prot.SignalProcessOptions) error {
+func (c *MockCore) SignalProcess(id string, pid int, options prot.SignalProcessOptions) error {
 	c.LastSignalProcess = SignalProcessCall{
```
beweedon: Add ID field to SignalProcessCall

service/gcs/core/mockcore/mockcore.go
```diff
@@ -159,11 +168,23 @@ func (c *MockCore) RegisterContainerExitHook(id string, exitHook func(oslayer.Pr
 
 // RegisterProcessExitHook captures its arguments, runs the given exit hook on
 // a process exit state with exit code 103, and returns a nil error.
-func (c *MockCore) RegisterProcessExitHook(pid int, exitHook func(oslayer.ProcessExitState)) error {
+func (c *MockCore) RegisterProcessExitHook(id string, pid int, exitHook func(oslayer.ProcessExitState)) error {
```
beweedon: Add ID field to RegisterProcessExitHookCall

service/gcs/core/mockcore/mockcore.go
beweedon: And update tests to match.

service/gcs/core/mockcore/mockcore.go
beweedon: And update tests.

service/gcs/runtime/runc/runc.go
```diff
@@ -528,8 +536,8 @@ func (c *container) startProcess(tempProcessDir string, hasTerminal bool, stdioS
 	}
 
 	var relay *stdio.TtyRelay
+	var master *os.File
```
beweedon: I'm probably missing something, but why move this to a higher scope?

service/gcs/stdio/stdio.go
```diff
@@ -4,10 +4,15 @@ import (
...
 	"sync"
+	"sync/atomic"
+	"syscall"
+	"unsafe"
 
 	"github.com/Microsoft/opengcs/service/gcs/transport"
 	"github.com/Sirupsen/logrus"
 	"github.com/pkg/errors"
+
+	"golang.org/x/sys/unix"
```
beweedon: Nit: To keep with the pattern we've started following (using goimports to perform formatting), this import should probably go with the above github.com imports.

jterry75: Done.

service/gcs/runtime/runc/runc.go
jterry75: Was a remnant before we used the TtyRelay to do the work. Put back now.

service/gcs/core/gcs/gcs.go
jterry75: ContainerID is not a required field. The HCS will not send any parameter and the unmarshal call from json will default "". I could be wrong but this seems safe.

service/gcs/stdio/stdio.go
```diff
@@ -100,15 +105,37 @@ func (s *ConnectionSet) Files() (_ *FileSet, err error) {
...
+		y      uint16
+	}
+
+	r.wg.Add(1)
+	defer r.wg.Done()
+	if atomic.LoadInt32(&r.closing) != 0 {
+		return errors.New("error resizing console pty is closed")
+	}
+
+	if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, r.pty.Fd(), uintptr(unix.TIOCSWINSZ), uintptr(unsafe.Pointer(&consoleSize{Height: height, Width: width}))); err != 0 {
```
jstarks: >Syscall [](start = 25, length = 7)

Probably you should make a standalone function in tty.go that calls this IOCTL that can be used or tested separately from the relay.

service/gcs/stdio/stdio.go
```diff
@@ -151,6 +178,13 @@ func (r *TtyRelay) Wait() {
 
 	// Wait for all users of stdioSet and master to finish before closing them.
 	r.wg.Wait()
+
+	// Given the expected use of wait we cannot increment closing before calling r.wg.Wait()
+	// or all calls to ResizeConsole would fail. However, by calling it after there is a very
+	// small window that ResizeConsole could still get an invalid Fd. We call wait again to
+	// enusure that no ResizeConsole call came in before actualling closing the pty.
+	atomic.StoreInt32(&r.closing, 1)
+	r.wg.Wait()
```
jstarks: >wg [](start = 3, length = 2)

I think you need to use a separate mutex unfortunately; according to the docs, you can't legally increment the wait group once Wait has been called:

> go doc sync.WaitGroup.Add
func (wg *WaitGroup) Add(delta int)
    Add adds delta, which may be negative, to the WaitGroup counter. If the
    counter becomes zero, all goroutines blocked on Wait are released. If the
    counter goes negative, Add panics.

    Note that calls with a positive delta that occur when the counter is zero
    must happen before a Wait. Calls with a negative delta, or calls with a
    positive delta that start when the counter is greater than zero, may happen
    at any time. Typically this means the calls to Add should execute before the
    statement creating the goroutine or other event to be waited for. If a
    WaitGroup is reused to wait for several independent sets of events, new Add
    calls must happen after all previous Wait calls have returned. See the
    WaitGroup example.

jterry75: Done.

service/gcs/stdio/stdio.go
jterry75: Done.

service/gcs/core/gcs/gcs.go
jterry75: Fixed.

service/gcs/core/mockcore/mockcore.go
jterry75: Fixed

service/gcs/core/mockcore/mockcore.go
jterry75: Fixed

service/gcs/stdio/stdio.go
```diff
@@ -151,6 +167,33 @@ func (r *TtyRelay) Wait() {
...
+
 	r.pty.Close()
+	r.closed = true
 	r.s.Close()
+
+}
+
+// ResizeConsole sends the appropriate resize to a pTTY FD
+// Synchronization of pty should be handled in the callers context.
+func ResizeConsole(pty *os.File, height, width uint16) error {
```
jstarks: >ResizeConsole [](start = 5, length = 13)

put this in tty.go

service/gcs/stdio/stdio.go
```diff
@@ -151,6 +167,33 @@ func (r *TtyRelay) Wait() {
...
+	type consoleSize struct {
+		Height uint16
+		Width  uint16
+		x      uint16
+		y      uint16
+	}
+
+	if pty == nil {
+		return errors.New("error resizing console for nil pty fd")
+	}
```
jstarks: Is this really necessary?

service/gcs/core/gcs/gcs.go
```diff
@@ -251,10 +243,6 @@ func (c *gcsCore) ExecProcess(id string, params prot.ProcessParameters, stdioSet
 			}
 			c.processCacheMutex.Unlock()
 			c.containerCacheMutex.Lock()
-			containerEntry.ExitStatus = state
-			for _, hook := range containerEntry.ExitHooks {
-				hook(state)
-			}
 			delete(c.containerCache, id)
```
beweedon: Why get rid of this?

service/gcs/stdio/stdio.go
jterry75: Done.

service/gcs/stdio/stdio.go
jterry75: Removed.

service/gcs/core/gcs/gcs.go
jterry75: This was a mistake for sure. Not sure how I missed that

service/gcs/bridge/bridge_test.go
```diff
@@ -613,10 +613,29 @@ var _ = Describe("Bridge", func() {
...
+						Height:    80,
+						Width:     80,
+					}
+				})
+				AssertNoResponseErrors()
+				AssertActivityIDCorrect()
+				It("should receive the correct values", func() {
+					Expect(callArgs.Pid).To(Equal(102))
+					Expect(callArgs.Height).To(Equal(uint16(80)))
+					Expect(callArgs.Width).To(Equal(uint16(80)))
```
beweedon: Oops, forgot to mention this on my last review: If we're getting rid of the `id==""` check in this PR, we should probably get rid of this test.

service/gcs/stdio/stdio.go
```diff
@@ -151,6 +164,12 @@ func (r *TtyRelay) Wait() {
...
 	// Wait for all users of stdioSet and master to finish before closing them.
 	r.wg.Wait()
+
+	r.m.Lock()
+	defer r.m.Unlock()
+
 	r.pty.Close()
+	r.closed = true
 	r.s.Close()
+
```
beweedon: Nit: unnecessary newline

service/gcs/bridge/bridge_test.go
jterry75: Removed.

service/gcs/stdio/stdio.go
jterry75: Removed.

service/gcs/core/gcs/gcs.go
```diff
@@ -38,39 +38,32 @@ type gcsCore struct {
 	containerCache map[string]*containerCacheEntry
 
 	processCacheMutex sync.RWMutex
-	// processCache stores information about processes which persists between
-	// calls into the gcsCore. It is structured as a map from pid to cache
-	// entry.
-	processCache map[int]*processCacheEntry
-
-	externalProcessCacheMutex sync.RWMutex
 	// externalProcessCache stores information about external processes which
```
jstarks: >externalProcessCache [](start = 4, length = 20)

comment needs updating

service/gcs/runtime/runtime.go
```diff
@@ -43,6 +43,7 @@ type Process interface {
 	Wait() (oslayer.ProcessExitState, error)
 	Pid() int
 	Delete() error
+	Tty() *stdio.TtyRelay
```
jstarks: >Tty [](start = 1, length = 3)

Maybe better to just expose the Resize method directly here? This seems to leak an implementation detail.

jstarks:
:clock1:

service/gcs/runtime/runtime.go
jterry75: I will open an issue to unify the runtime Process and the oslayer Process interfaces. The issue is that for container/external processes we store them quite differently and the cache cant support both. We really should be thinking about them the same with a single Process interface from the GCS perspective. Will this work for now?

jterry75:
Anymore feedback here? Would like to get this checked in

PR #85, opened 2017-08-01
Added readdir and fixed typo in omitempty

gupta-ak:
With this, I think all of the docker side stuff can be implemented.


soccerGB:
Changes look good

PR #86, opened 2017-08-01
(Final!) refresh of client

jhowardmsft:

Hopefully the last refresh before I kill jhowardmsft/opengcs and transition docker over to Microsoft/opengcs instead later this week. Note it's dependent on 34272 which is the rest of docker moving over to lower-case sirupsen\logrus, imminently about to be merged into master there.

### DIFF
--- a/client/config.go
+++ b/client/config.go
@@ -11,7 +11,7 @@ import (
 	"strings"
 
 	"github.com/Microsoft/hcsshim"
-	"github.com/Sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 )
 
 // Mode is the operational mode, both requested, and actual after verification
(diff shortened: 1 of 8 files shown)


client/config.go
```diff
@@ -160,8 +160,8 @@ func (config *Config) GenerateDefault(options []string) error {
 	return nil
 }
 
-// validate validates a Config structure for starting a utility VM.
-func (config *Config) validate() error {
+// Validate validates a Config structure for starting a utility VM.
+func (config *Config) Validate() error {
```
jterry75: JW why our exporting this now but I see no new callers? Does Docker call this?

jterry75:
You might have to git mv the vendor/github.com/Sirupsen folder to be lower case but not sure

jhowardmsft:
Yeah, that's for the server side of opengcs, not the client. There's no vendoring in the client library - the user of the library (docker being the only one currently) is responsible for vendoring which has already been updated :)

client/config.go
jhowardmsft: There is in a PR about to be submitted.

jterry75:
Oh. Yuck... I will submit a change for the service side.

PR #87, opened 2017-08-02
Update to lowercase sirupsen/logrus

jterry75:


jterry75:
Matching #86 on the service side.

jstarks:
LGTM

gupta-ak:
Guess the new logrus needs `golang.org/x/crypto/ssh/terminal` so could you add that in?

PR #94, opened 2017-08-04
remotefs: Added sync to writes and OpenFile

gupta-ak:
Added a `sync` for writes, so that writes from `docker cp`  will always persistent to the container.

Also added a `OpenFile` command to support cases where docker calls `Open/OpenFile`. Since the launch process architecture makes it hard for `remotefs` to keep state, the `OpenFile` command is mainly a permission check. After calling `OpenFile`, docker can call `ReadFile` to get the file on the host, use it as a normal `*os.File` and then do `WriteFile` to send the file to the VM to write to the VHDX.


service/gcsutils/remotefs/utils.go
```diff
@@ -146,3 +147,9 @@ func WriteTarOptions(w io.Writer, opts *archive.TarOptions) error {
 
 	return nil
 }
+
+// Sync syncs the filesystem so that all writes buffered in the kernel are flushed
+// to disk.
+func Sync() error {
+	return exec.Command("sync").Run()
```
jstarks: This is a pretty heavy hammer. Can't we implement this just to fsync the files that have been modified?

service/gcsutils/remotefs/remotefs_os.go
```diff
@@ -239,7 +258,45 @@ func Mkfifo(in io.Reader, out io.Writer, args []string) error {
...
-	return unix.Mkfifo(args[0], uint32(perm))
+
+	if err := unix.Mkfifo(args[0], uint32(perm)); err != nil {
+		return err
+	}
+	return Sync()
+}
+
+// OpenFile works like os.OpenFile. Since the GCS process calling structure
+// does not enable us to keep state, OpenFile doesn't return a handle to the file.
```
jstarks: :(. What do we need to do to improve this?

service/gcsutils/remotefs/utils.go
gupta-ak: Yeah, I can try with `fsync`

service/gcsutils/remotefs/remotefs_os.go
gupta-ak: I'm not sure. I guess a workaround can be for docker to call `remotefs OpenFile` and then have a small protocol for reading/writing/seeking/closing through `stdin` & `stdout`

gupta-ak:
@jstarks Updated

1) Removed the `Sync()` stuff because it's not necessary. `umount` already does a `sync`, so the data should be flushed if `docker` calls `graphdriver.Put()`. The reason why I wasn't seeing it being synced because I forgot to wait for `umount` to complete.
2) I changed the OpenFile implementation to have a "mini fileserver" protocol in to support `*os.File Read/Write/Close`, so the client now keeps a handle to the `remotefs` process and issues a `Close command` when they're done with the file.

Issue #78, opened 2017-07-25
/etc/resolv.conf not getting populated in ubuntu image


jstarks:
```docker run --rm ubuntu cat /etc/resolv.conf```

No output. This breaks DNS resolution inside containers.

**Comment by rn** [comment]
I made some changes to this a week or so ago via: https://github.com/Microsoft/opengcs/pull/63 I also left this rather length comment: https://github.com/Microsoft/opengcs/blob/master/service/gcs/core/gcs/gcs.go#L183

I did test DNS/`resolv.conf` at the the time and it worked with the changes in my PR (it didn't before for a variety of reasons, including that `/etc/resolv.conf` was written in an incorrect format). So maybe something else changed in the meantime...

**Comment by gupta-ak** [comment]
It shows up in both the `ubuntu` and `busybox` image. In the `busybox` image, it is actually populated, but in the `ubuntu` image it is not. @beweedon  any idea why?

**Comment by gupta-ak** [comment]
Ah, I read Rolf's comment and checked the `ubuntu` image. Turns out one of the `ubuntu` layers has an empty `resolv.conf` so `overlay` overrides it.

**Comment by rn** [comment]
yeah, I just noticed that as well.

If you look at the `json.conf` of a container on Linux you see (this is from a `docker run --rm -ti busybox /bin/sh` and then a `runc state <id>` which gives you directory of the bundle which contains the `config.json`):
```
[...]
{"destination":"/etc/resolv.conf","type":"bind","source":"/var/lib/docker/containers/
8a46042f5c5d7672895ad0c5357c9819bdfac3d939767c52010d301b2fc7c85b/resolv.conf","options":["rbind","rprivate"]},{"destination":"/etc/hostname","
type":"bind","source":"/var/lib/docker/containers/8a46042f5c5d7672895ad0c5357c9819bdfac3d939767c52010d301b2fc7c85b/hostname","options":["rbind
","rprivate"]},{"destination":"/etc/hosts","type":"bind","source":"/var/lib/docker/containers/8a46042f5c5d7672895ad0c5357c9819bdfac3d939767c52
010d301b2fc7c85b/hosts","options":["rbind","rprivate"]}
[...]
```
so this bind mounts a `resolv.conf`, `hostname`, and `hosts` into the container, effectively overwriting whatever is in the image. These bind mounts are not present in the `mounts` section of the `config.json` pass from the windows side to the utility VM. Hence my comment about bind mounts.

beweedon:
[...]

PR #95, opened 2017-08-04
Changing base layer order

jterry75:
Partially Resolves: #78

### DIFF
--- a/service/gcs/core/gcs/storage.go
+++ b/service/gcs/core/gcs/storage.go
@@ -13,9 +13,9 @@ import (
 
 	"github.com/Microsoft/opengcs/service/gcs/oslayer"
 	"github.com/Microsoft/opengcs/service/gcs/prot"
-	"github.com/sirupsen/logrus"
 	oci "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


jterry75:
I have tested this e2e and verified that something such as:

`docker run --rm ubuntu cat /etc/resolv.conf` does return the correct name servers

service/gcs/core/gcs/storage.go
```diff
@@ -311,23 +311,23 @@ func (c *gcsCore) mountLayers(id string, scratchMount *mountSpec, layers []*moun
 	// Mount the layer devices.
 	layerPaths := make([]string, len(layers)+1)
 	for i, layer := range layers {
-		layerPath := fmt.Sprintf("%s%d", layerPrefix, i)
+		layerPath := fmt.Sprintf("%s%d", layerPrefix, i+1)
```
soccerGB: Interesting change.
Was the old code simply incorrect?

jterry75:
Oops. Shifting the layer names broke the unit tests.

We actually could leave that part alone. I did that to maintain consistency with the array order in  `layerPaths` given that we shifted them all by `1`. If that is confusing I can put it back however that will  mean *layer0* is actually *layer1* in the overlay. Do you care one way over another?

Or I can just update the unit tests to read from the correct locations.

gupta-ak:
I think it's fine if `layerPath := fmt.Sprintf("%s%d", layerPrefix, i+1)` gets changed to a `i` instead. The `layer[0]` is bascially as fake layer for our networking stuff, so it makes sense that `layer0` = first layer of real image overlay = `layer[1]`.

gupta-ak:
It's because the baselayer was originally the bottom most read-only layer containing the `resolv.conf` for networking. If any image, like `ubuntu`, had the `/etc/resolv.conf` file in one of the layers, it would override our filled `resolv.conf` with whatever the layer had, breaking networking. Since now it's at the top, the correct `resolv.conf` won't get overwritten. See #78 for further discussion.

soccerGB:
Ah, I didn't connect those two topics.  Thanks for the explanation

jterry75:
Re-verified `docker run --rm ubuntu cat /etc/resolv.conf` with `layerPath := fmt.Sprintf("%s%d", layerPrefix, i)` and this passes unit tests again. Will work on another fix for `/sbin/udhcpc_config.script` separately

Issue #102, opened 2017-08-08
Update runC tests to include stdio and non-sleep processes


beweedon:
Due to a bug I wasn't able to track down, container processes would occasionally exit at random in the test suite. This couldn't be reproed in any other environment than the tests. A workaround was to use only `sleep` processes in the runC tests. However, some combination of GCS and runC updates seems to have fixed the bug. The tests should now be expanded to include other types of processes than `sleep`.

PR #106, opened 2017-08-10
Expanded runC tests

beweedon:
Add some runc tests. These tests couldn't be added previously because of a bug that only reproed in tests which caused about every 11th container process to exit. However, some combination of runc and GCS changes seem to have squashed that bug, so a greater number and variety of tests can now be added.

This PR also changes when the call to wait for runc occurs on create/exec, and reimplements MockConnection to use Unix sockets rather than os.Pipes (which allows for implementation of the File() method on transport.Connection).

Deals with issue #102


service/gcs/transport/mock.go
```diff
@@ -15,43 +19,54 @@ type MockTransport struct {
 
 // Dial ignores the port, and returns a MockTransport struct.
 func (t *MockTransport) Dial(_ uint32) (Connection, error) {
-	csr, csw := io.Pipe()
-	scr, scw := io.Pipe()
+	sockDir, err := ioutil.TempDir("", "mocktport")
```
jterry75: Why can we do here with unix sockets we couldnt with pipes? I am confused why we need a "more full" moc impl

beweedon: os.Pipe doesn't support the File() method. This is a problem for the runc wrapper code, which relies on this method existing so that it can dup the file descriptors into the runc process. Without this change, the runc tests fail if they try to do anything with stdio.

jterry75: Ah ok.

gupta-ak:
I'm a bit confused on the reasoning for the more complicated MockTransport. If you just need the File() method, couldn't you just provide a trivial implementation with syscall.Dup and os.NewFile?

service/gcs/transport/mock.go
beweedon: Also, oops, sorry, meant io.Pipe.

jstarks:
> 			cmd.Wait()

This seems unnecessary now...

---
Refers to: service/gcs/runtime/runc/runc.go:546 in 5d6bc77. [](commit_id = 5d6bc77585963b4ea9a7ca636ba9f393b7761a20, deletion_comment = False)

beweedon:
I'm not sure I understand. io.Pipe doesn't produce anything with a file descriptor as far as I can tell. If I used os.Pipe, I still wouldn't know which file (read or write) to return for a given call to File(). Sorry, I'm not the most well-versed on the nuances of files on Linux.

service/gcs/transport/mock.go
```diff
@@ -15,43 +19,54 @@ type MockTransport struct {
...
+		return nil, err
+	}
+	name := filepath.Join(sockDir, "mocktport.sock")
+	addr, err := net.ResolveUnixAddr("unix", name)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to resolve unix socket address %s", name)
+	}
+
+	// Set up server connection.
+	listener, err := net.ListenUnix("unix", addr)
```
jstarks: >ListenUnix [](start = 22, length = 10)

It would probably be simpler to use socketpair to get a pair of connected UNIX sockets.

jstarks: i.e. syscall.Socketpair + net.FileConn

---
In reply to: [132545253](https://github.com/Microsoft/opengcs/pull/106#discussion_r132545253) [](ancestors = 132545253)

jstarks:
:clock1:

service/gcs/transport/mock.go
beweedon: Oh, I didn't know about that function! I'll make that change!

beweedon:
bump

service/gcs/transport/mock.go
```diff
@@ -15,43 +17,35 @@ type MockTransport struct {
 
 // Dial ignores the port, and returns a MockTransport struct.
 func (t *MockTransport) Dial(_ uint32) (Connection, error) {
-	csr, csw := io.Pipe()
-	scr, scw := io.Pipe()
+	fds, err := syscall.Socketpair(syscall.AF_UNIX, syscall.SOCK_STREAM, 0)
```
jterry75: I don't know the expectation here but can this be leaked if any of the calls fail?

beweedon: You're totally right. I can fix that now!

beweedon: OK, the fix is up.

beweedon:
Fixing what I think is the cause to a periodic test failure.

Issue #103, opened 2017-08-08
Test Bridge error paths


beweedon:
At the moment, the bridge's error paths are not being tested. They should.

jterry75:
This has been fully completed a while back but forgot to mention it in the 2nd review #125 too. Closing.

Issue #101, opened 2017-08-08
Better error and panic recovery in GCS


beweedon:
On a panic or on an error, the GCS shouldn't crash, exit, or otherwise stop responding. It should do a better job of recovering from errors like this.

beweedon:
Not going to do this, in favor of fail-fast.

PR #107, opened 2017-08-10
Proper error recovery in the bridge

beweedon:
The bridge now handles errors and panics better. This occurs in two layers:

1. If one of the bridge's methods (e.g. `createContainer`) panics, this will now be turned into an error by the method. From the bridge `loop`'s point of view the behavior hasn't changed, but the methods have to do a bit of extra work to turn the panic into an error.
2. If the bridge `loop` panics or errors, it will return the error up to `CommandLoop`. `CommandLoop` will then retry on errors until `Kill()` is called on the bridge. This allows the bridge to listen for the next message, even if it hasn't properly responded to the last one. This allows for things like opening up a shell into the UVM even after a bridge error occurs (assuming the container isn't in an invalid state on the HCS end).

Addresses issues #101 and #103

### DIFF
--- a/service/gcs/bridge/bridge_test.go
+++ b/service/gcs/bridge/bridge_test.go
@@ -123,6 +123,46 @@ var _ = Describe("Bridge", func() {
 			)
 			BeforeEach(func() {
 				messageType = prot.ComputeSystemCreateV1
+				settings = prot.VMHostedContainerSettings{
+					Layers:          []prot.Layer{prot.Layer{Path: "0"}, prot.Layer{Path: "1"}, prot.Layer{Path: "2"}},
+					SandboxDataPath: "3",
+					MappedVirtualDisks: []prot.MappedVirtualDisk{
+						prot.MappedVirtualDisk{
+							ContainerPath:     "/path/inside/container",
+							Lun:               4,
+							CreateInUtilityVM: true,
+							ReadOnly:          false,
+						},
+					},
+					NetworkAdapters: []prot.NetworkAdapter{
+						prot.NetworkAdapter{
+							AdapterInstanceID:  "00000000-0000-0000-0000-000000000000",
+							FirewallEnabled:    false,
+							NatEnabled:         true,
+							AllocatedIPAddress: "192.168.0.0",
+							HostIPAddress:      "192.168.0.1",
+							HostIPPrefixLength: 16,
+							HostDNSServerList:  "0.0.0.0 1.1.1.1 8.8.8.8",
+							HostDNSSuffix:      "microsoft.com",
+							EnableLowMetric:    true,
+						},
+					},
+				}
+				settingsBytes, err := json.Marshal(settings)
+				Expect(err).NotTo(HaveOccurred())
+				message = prot.ContainerCreate{
+					MessageBase: &prot.MessageBase{
+						ContainerID: containerID,
+						ActivityID:  activityID,
+					},
+					ContainerConfig: string(settingsBytes),
+					SupportedVersions: prot.ProtocolSupport{
+						MinimumVersion:         "V3",
+						MaximumVersion:         "V3",
+						MinimumProtocolVersion: prot.PvV3,
+						MaximumProtocolVersion: prot.PvV3,
+					},
+				}
 			})
 			JustBeforeEach(func() {
 				response = prot.ContainerCreateResponse{}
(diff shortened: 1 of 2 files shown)


service/gcs/bridge/bridge.go
```diff
@@ -34,6 +34,11 @@ type bridge struct {
 	// message data is kept together when writing from multiple go routines
 	// simultaneously.
 	writeLock sync.Mutex
+
+	// killed and killedLock determine if the bridge has been killed.
+	// killedLock must be held when reading or writing killed.
+	killed     bool
+	killedLock sync.Mutex
```
jterry75: We don't need a mutex just to check a state variable. Use the atomic types

service/gcs/bridge/bridge.go
```diff
@@ -45,23 +50,57 @@ func NewBridge(tport transport.Transport, coreint core.Core) *bridge {
 	}
 }
 
+// panicToError converts a panic return value into an error value.
+func panicToError(r interface{}) error {
```
jterry75: I'm new to go but when does a panic occur in go where its a good thing to recover. Panic's aren't the error model they are the exception model. It seems like panic's are likely always issues in our stack we need fixed. Am I wrong about that?

service/gcs/bridge/bridge.go
```diff
@@ -45,23 +50,57 @@ func NewBridge(tport transport.Transport, coreint core.Core) *bridge {
...
+	return errors.Errorf("PANIC: %s", r)
+}
+
+// Kill closes the bridge's command connection and sets killed to true. In this
+// way, any outstanding command reads should fail, and the retry loop should
+// exit. This is useful when testing, where we don't want a retry loop running
+// constantly in the background for every test to slow them all down.
+func (b *bridge) Kill() {
+	b.commandConn.Close()
+	b.killedLock.Lock()
```
jterry75: again. `atomic.StoreInt32(&killed, 1)`

service/gcs/bridge/bridge.go
```diff
@@ -45,23 +50,57 @@ func NewBridge(tport transport.Transport, coreint core.Core) *bridge {
...
+	if b.commandConn, err = b.createAndConnectCommandConn(); err != nil {
 		return err
 	}
 	defer b.commandConn.Close()
 
+	// Keep retrying the loop on error until Kill has been called.
+	for {
+		err = b.loop()
+		if err != nil {
+			b.killedLock.Lock()
```
jterry75: And finally
`
if atomic.LoadInt32(&killed) != 0 {
    break;
}
`

jstarks:
I don't agree with the principle of this change. We should not be trying to recover panics. If there is a panic, it means there is a bug and we have to fix it. Recovering often means leaving the process in an indeterminate state.

________________________________

beweedon:
John and Justin, I agree with you both that panics are indicative of a bug in the GCS that needs to be fixed. That said, when testing, it's frustrating to encounter a panic and have it bring the whole GCS down. This can make it more difficult to root cause the problem, especially without easy access to something like porthole. Panics, like some error values we might encounter, are indicative of bugs in the GCS. But I think there's still value to handling these bugs gracefully.

As for leaving the process in an indeterminate state, correct me if I'm wrong, but because panics still call all the defers in the stack until they recover, cleanup operations (which we still in general need to be better at handling on error, but that's another issue) can still be run.

service/gcs/bridge/bridge.go
beweedon: I considered that, but preferred the strong typing of bools over ints. That said, I don't really have a strong opinion either way and I think atomics are more performant at least in C++, so I'll make the change :)

jstarks:
There is no way to know whether recovering is going to keep the process in a good state. Some code uses defer to undo things on failure, some code manually cleans up resources after checking err values, and some code is not possible to undo in the case of a panic. There is no concept of panic safety in Go like there is a concept of exception safety in C++.

In everything we do, we must follow the fail fast philosophy -- if there is a logic bug or fatal error, we must terminate the process. Doing otherwise hides security bugs, data loss bugs, and a variety of weird and unpredictable behavior.

If the problem is that it's too hard to determine what went wrong when there is a panic, then we need to find another way to ensure that the GCS logs and panic output make it back to the host. For example, maybe we need to have GCS try to open a well-known hvsocket on startup to use for stderr. Then we can provide a small tool on the host that binds to that socket that can be used to make debugging easier.

gupta-ak:
How I've been debugging GCS panic is redirecting gcs stdout/stderr to a persistent file on disk (using VHD boot). I imagine something similar could be to hook up the GCS's stdout/err to well known hvsockets.

Another option is to get the serial console stuff working and then just read from them. This would also help with debugging kernel panics, which can't really be debugged at the moment.

jstarks:
Yeah, the serial console is probably the most straightforward path. We could have two serial ports, one for the Linux console and interactive shell, and a second one for the GCS log (and possibly service VM tool logs).

beweedon:
That's fair. We can explore other avenues for getting panic information back. If we want to go for fail fast, then I should probably also remove the retry logic for the bridge loop, since any errors in the loop itself are likely to either be major errors in the GCS code or malformed data sent from the HCS, both of which are difficult to figure out how to properly recover from.

If this seems cool, I can just redo the commits to only contain the error handling tests.

jstarks:
Sounds good.

beweedon:
bump

service/gcs/bridge/bridge.go
```diff
@@ -334,21 +334,37 @@ func (b *bridge) runExternalProcess(message []byte) (*prot.ContainerExecuteProce
 	return response, nil
 }
 
-func (b *bridge) waitOnProcess(message []byte, header *prot.MessageHeader) (*prot.ContainerWaitForProcessResponse, error) {
-	response := &prot.ContainerWaitForProcessResponse{MessageResponseBase: newResponseBase()}
+func (b *bridge) waitOnProcess(message []byte, header *prot.MessageHeader) (response *prot.ContainerWaitForProcessResponse, err error) {
```
jterry75: What is the real issue here? Im wondering if we shouldnt take this case in favor of my async bridge change which removes exit hooks altogether.

beweedon: The issue is that in theory, if RegisterExitHook fails, it could still register the exit hook and thus two responses (both an error and the one specified by the exit hook) could be sent. This would never occur if RegisterExitHook behaves properly, but I put this change in mainly to not send nonsense back to the HCS if it does not behave properly. I'm fine with not making this change, however, especially if your async bridge changes remove the necessity for it.

beweedon:
Bump

jterry75:
Bump. It would be helpful to merge this or the async bridge request so we can rebase and include the set of changes.

jterry75:
Given this is a test only change and its passing. I am going to merge so I can make the updates to async bridge PR.

PR #110, opened 2017-08-10
client: Add travis CI verification

jhowardmsft:

Adds CI for the client library
- Ensures all files are correctly formatted
- Ensures all files are linted
- Ensures the library builds

In the future I'll add some unit tests too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,15 @@ services:
 install: ./service/gcs/setup_test_env
 
 script:
+  # Server-side checks
   - cd service && make
   - go test ./gcsutils/...
   - cd gcs && sudo env "PATH=$PATH" ginkgo -r -race -keepGoing
+
+  # Client-side checks
+  - cd ../../client
+  - go get -u github.com/golang/lint/golint 
+  - go get github.com/GeertJohan/fgt
+  - bash test/.checkgocode.sh
+  - go get -d github.com/Microsoft/hcsshim github.com/Microsoft/go-winio
+  - export GOOS="windows" && go build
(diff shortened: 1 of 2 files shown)


jhowardmsft:
Ready for review. If you're wondering why using fgt, see the comment in the code. If you're wondering why not put it directly in the .yml, I tried, but CI always succeeds, even if golint/gofmt fail. This however does work. And will keep the client code clean :)

gupta-ak:
I do see a `-set_exit_status` flag on golint:
```
PS C:\Users\akagup\go\src\github.com\Microsoft\opengcs> golint.exe --help
Usage of C:\Users\akagup\go\bin\golint.exe:
        golint [flags] # runs on package in current directory
        golint [flags] [packages]
        golint [flags] [directories] # where a '/...' suffix includes all sub-directories
        golint [flags] [files] # all must belong to a single package
Flags:
  -min_confidence float
        minimum confidence of a problem to print it (default 0.8)
  -set_exit_status
        set exit status to 1 if any issues are found
```

.travis.yml
```diff
@@ -16,6 +16,15 @@ services:
...
 
 script:
+  # Server-side checks
   - cd service && make
   - go test ./gcsutils/...
   - cd gcs && sudo env "PATH=$PATH" ginkgo -r -race -keepGoing
+
+  # Client-side checks
+  - cd ../../client
+  - go get -u github.com/golang/lint/golint
```
soccerGB: I recommend move those two "go get" operations into ./service/gcs/setup_test_env, which has all the test prep stuffs.

PR #111, opened 2017-08-10
Kernelconfig refactoring and add missing patches

miguelinux:
To get a similar directory as linuxkit kernel project and to be easy to end user to create a custom linux kernel the following changes were made.

- Refactoring the kernelconfig directory to be similar to the linuxkit project
- Add the Hyper-V vsock missing patches. Now the user don't need to cherry-pick them.
- Organize patch set to be easy to apply
- Consolidate documentation in one file and explain how to apply patches


soccerGB:
In general, the changes look pretty good. The only comment I have is about the adding of those Hyper-V vsock related patches into this repos. We were trying not to copy any patches into this opengcs repo, that's why we added specifically  "cherry pick instructions" for it.  There are pro and con in doing. Before the on-going Hyper-V vsock activities died down (they are still chasing some issues, it might be a better idea to keep it separate, not copying down patches to this repro.  Dose that make sense to you?

miguelinux:
@soccerGB Ok, Let me remove them from these patches and keep only the instructions.

Issue #98, opened 2017-08-08
Check file permissions throughout GCS code


beweedon:
Sometimes the GCS opens a file or sets the permissions on a file. We should check to make sure the permissions being used in each instance are what we want, and change them if not.

PR #112, opened 2017-08-11
Change permissions of etc directory from 0700 to 0755

beweedon:
This fixes an issue experienced in the ubuntu image where apt-get needs
the /etc directory to be executable to perform an nslookup. When we
switch to using bind mounts specified by docker to get resolv.conf into
the container namespace, we might be able to avoid specifying the
permission bits altogether.


beweedon:
#98

PR #113, opened 2017-08-15
Fix nil deref when closing ConnectionSet on error

beweedon:
This bug caused the GCS to crash with a panic when connecting to stdio
sockets failed in any way.

### DIFF
--- a/service/gcs/bridge/connection.go
+++ b/service/gcs/bridge/connection.go
@@ -4,12 +4,11 @@ import (
 	"encoding/binary"
 	"io"
 
-	"github.com/sirupsen/logrus"
-	"github.com/pkg/errors"
-
 	"github.com/Microsoft/opengcs/service/gcs/prot"
 	"github.com/Microsoft/opengcs/service/gcs/stdio"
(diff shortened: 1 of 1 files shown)


jhowardmsft:
Nice! LGTM

jterry75:
Why use named return values in this case if you are going to use a local?

beweedon:
The named return values are for the defer error checking. Doing it this way, when you return an error, the named err return value is guaranteed to be set, and can be used by the defer function.

gupta-ak:
think hes means the 's': `(s *stdio.ConnectionSet, err error)`

beweedon:
If we specify one named return value (`err`), we have to specify the other (`s`) or else it's a compile error.

gupta-ak:
`(_ *stdio.ConnectionSet, err error)` isn't a compile error, since the name for stdio.ConnectionSet still specified as _, the value is just ignored. If you take out the dash, like `(*stdio.ConnectionSet, err error)`, then it would be a compile error.

beweedon:
Oh, I didn't know the underscore was valid syntax here! Sure, I'll make that change!

PR #116, opened 2017-08-18
Implements the async bridge loop

jterry75:
Implements the async bridge
Implements the bridge multiplexer
Removes any waiting between commands in favor of a channel writer per request
Fixes the "RegisterExitHook" pattern in favor of a single channel writer on an async goroutine
Temporarily disables the bridge_test suite put puts in place bridge unit tests. These will be moved when the handler funcs are pulled off of the bridge in the second change.


jterry75:
I have verified both hcs based and docker based scenarios.

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+	coreint core.Core
+
+	// testing hook to close the bridge ListenAndServe() method
+	quitChan chan bool
+}
+
+// AssignHandlers creates and assigns the appropriate bridge
+// events to be listen for and intercepted on `mux` before forwarding
+// to `gcs` for handling.
+func (b *Bridge) AssignHandlers(mux *Mux, gcs core.Core) {
```
jterry75: This is to be moved into the gcs folder but to minimize the change to make review easier I did it this way.

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+	// ComputeSystemCreate) over.
+	commandConn transport.Connection
+
+	responseChan chan bridgeResponse
+
+	// Core - TODO: Remove this and use the mux!
+	coreint core.Core
+
+	// testing hook to close the bridge ListenAndServe() method
+	quitChan chan bool
```
jterry75: Dont think it matters now but if we used `signal.Notify` for `ctrl+c` we could actually gracefully shut down the bridge listener in the future

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
-				return errors.Errorf("invalid response type: %T", response)
+			if _, err := b.commandConn.Write(responseBytes); err != nil {
+				responseErrChan <- errors.Wrap(err, "bridge: failed writing message payload")
+				continue
 			}
+			logrus.Infof("bridge: response sent: '%s' to HCS\n", responseBytes)
 		}
+	}()
+	// If we get any errors. We return from Listen and shutdown the bridge connection.
+	// TODO: We should likely just log a generic response and keep listening?
```
jterry75: Looking for feedback here. Is it right to stop the GCS in this case?

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+			Type: prot.ComputeSystemNotificationV1,
+			ID:   0,
+		},
+		response: n,
 	}
+	b.responseChan <- resp
 }
 
-func (b *bridge) createContainer(message []byte) (*prot.ContainerCreateResponse, error) {
+func (b *Bridge) createContainer(w ResponseWriter, r *Request) {
```
jterry75: Again, all of these functions will move to the gcs folder

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+		w.Error(err)
+		return
 	}
 
-	exitHook := func(state oslayer.ProcessExitState) {
-		if err := b.sendExitNotification(id, response.ActivityID, state); err != nil {
+	response.SelectedProtocolVersion = prot.PvV3
+	w.Write(response)
+
+	go func() {
```
jterry75: No more exit hooks!

service/gcs/core/gcs/gcs.go
```diff
@@ -56,25 +56,23 @@ func NewGCSCore(rtime runtime.Runtime, os oslayer.OS) *gcsCore {
 // containerCacheEntry stores cached information for a single container.
 type containerCacheEntry struct {
 	ID                 string
-	ExitStatus         oslayer.ProcessExitState
```
jterry75: I'm not committed to using `exitCode` vs `ExitStatus` but nobody ever uses it for anything other than the code so it seemed easier to only cache the code

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+	const commandPort uint32 = 0x40000000
+
+	var err error
+	b.commandConn, err = b.Transport.Dial(commandPort)
+	if err != nil {
+		return errors.Wrap(err, "bridge: failed creating the command Connection")
+	}
+	logrus.Info("bridge: successfully connected to the HCS via HyperV_Socket\n")
+
+	requestChan := make(chan *Request)
```
gupta-ak: Don't you want buffered channels for the request / response? The send will block until the other goroutine reads from it, so it's still serialized a bit.

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
-			logrus.Info("received from HCS: ComputeSystemResizeConsoleV1")
-			response, err = b.resizeConsole(message)
+			logrus.Infof("bridge: read message '%s'\n", message)
+			requestChan <- &Request{header, message}
+		}
+	}()
+	// process each bridge request async and create the response writer
+	go func() {
+		for req := range requestChan {
+			go func(r *Request) {
```
beweedon: Do you need to pass in `req` as a parameter, or can you just use the fact that `func` is a closure to use that variable?

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+					respChan: b.responseChan,
+				}
+				b.Handler.ServeMsg(wr, r)
+				if !wr.respWritten {
+					logrus.Errorf("bridge: request: ID: %d, Type: %d failed to write a response.\n", r.Header.ID, r.Header.Type)
+				}
+			}(req)
+		}
+	}()
+	// process each bridge response sync this is valid for request/response and publish workflows
```
beweedon: nit: "that is"?

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+	defer close(b.responseChan)
+	defer close(responseErrChan)
+	defer close(b.quitChan)
+
+	// receive bridge requests and schedule them to be processed.
+	go func() {
+		for {
+			header := &prot.MessageHeader{}
+			if err := binary.Read(b.commandConn, binary.LittleEndian, header); err != nil {
+				requestErrChan <- errors.Wrap(err, "bridge: failed reading message header")
```
beweedon: Or maybe `break` since we want to exit on errors anyway?

service/gcs/bridge/bridge.go
beweedon: I think this falls under the fail-fast discussion we had before. There are probably some specific connection errors it makes sense to retry, but for most errors (the connection is closed, the HCS sends an invalid message, we send an invalid message), it seems like something is probably pretty wrong/unrecoverable, and we should just exit rather than trying to recover. That isn't to say we shouldn't not retry/try to recover on certain errors, but I feel like the default behavior should just be to exit. Does that make sense?

service/gcs/bridge/bridge.go
beweedon: And the JSON unmarshaling will stay in the bridge, right?

service/gcs/bridge/bridge_unit_test.go
```diff
@@ -0,0 +1,619 @@
+package bridge
+
+import (
+	"encoding/binary"
+	"encoding/json"
+	"io"
+	"strings"
+	"sync"
+	"testing"
```
beweedon: Are you planning on using `testing` in the long term for these tests, or switching to ginkgo?

service/gcs/bridge/bridge_unit_test.go
```diff
@@ -0,0 +1,619 @@
...
+	}
+	if len(base.ErrorRecords) != 1 {
+		t.Error("The default handler did not set an error record.")
+	}
+	if !strings.Contains(base.ErrorRecords[0].Message, "bridge: function not supported") {
+		t.Error("The default handler did not return the not supported message")
+	}
+}
+func Test_Bridge_Mux_Handler_NotAdded_Default(t *testing.T) {
+	//
```
beweedon: nit: I think the pattern of having empty // lines before and after a comment is pretty uncommon in Go (at least from what I've seen).

service/gcs/bridge/bridge_unit_test.go
```diff
@@ -0,0 +1,619 @@
...
+	// Testing specifically that if we have a bridge with no handlers that
```
beweedon: nit: "that for"?

service/gcs/core/gcs/gcs.go
```diff
@@ -578,6 +528,34 @@ func (c *gcsCore) ResizeConsole(pid int, height, width uint16) error {
 	return p.Tty.ResizeConsole(height, width)
 }
 
+// WaitContainer waits for a container to complete and returns its exist code.
```
beweedon: nit: "exit"

service/gcs/core/gcs/gcs_test.go
```diff
@@ -1045,55 +1045,59 @@ var _ = Describe("GCS", func() {
...
+					exitCode, err = coreint.WaitContainer(containerID)
 				})
-				Context("the container has already been created", func() {
-					BeforeEach(func() {
+				Context("container does not exist", func() {
+					It("should produce errors", func() {
+						Expect(err).To(HaveOccurred())
+					})
+				})
+				Context("container does exit", func() {
```
beweedon: nit: this is my bad, but "exist"

service/gcs/core/gcs/gcs_test.go
```diff
@@ -1045,55 +1045,59 @@ var _ = Describe("GCS", func() {
...
-					Context("the process has already been started", func() {
-						BeforeEach(func() {
-							pid, err = coreint.ExecProcess(containerID, initialExecParams, fullStdioSet)
-							Expect(err).NotTo(HaveOccurred())
-						})
-						It("should not produce an error", func() {
-							Expect(err).NotTo(HaveOccurred())
-						})
+				})
+				Context("process does exit", func() {
```
beweedon: nit: my bad, "exist"

service/gcs/core/gcs/gcs_test.go
```diff
@@ -1045,55 +1045,59 @@ var _ = Describe("GCS", func() {
...
 				)
 				JustBeforeEach(func() {
-					err = coreint.RegisterProcessExitHook(pid, func(oslayer.ProcessExitState) {})
+					exitCode, err = coreint.WaitProcess(pid)
 				})
-				Context("the container has already been created", func() {
-					BeforeEach(func() {
-						err = coreint.CreateContainer(containerID, createSettings)
-						Expect(err).NotTo(HaveOccurred())
+				Context("process does not exit", func() {
```
beweedon: "exist"

service/gcs/core/mockcore/mockcore.go
```diff
@@ -186,3 +161,19 @@ func (c *MockCore) ResizeConsole(pid int, height, width uint16) error {
 
 	return nil
 }
+
+// WaitContainer captures its arguments and returns a nil error.
+func (c *MockCore) WaitContainer(id string) (int, error) {
+	c.LastWaitContainer = WaitContainerCall{
+		ID = id,
+	}
+	return -1, nil
```
beweedon: note: the return value will have to be changed to use `behaviorResult` after my bridge testing change is merged in.

service/gcs/bridge/bridge.go
jterry75: Yes that would also be an exit case.

service/gcs/bridge/bridge.go
jterry75: Deal. Will remove the comment. If we need to special case any errors we can do so later.

service/gcs/bridge/bridge.go
jterry75: No. I think that fundamentally the `bridge` is like `http` it manages reading the headers and calling the appropriate handlers providing the `header` and `body`. Its up to the handler to do something with that `body` and it knows what types of messages are allowed and such. Do you disagree with that?

service/gcs/bridge/bridge.go
jterry75: I guess *technically* your correct however the `requestChan` is immediately read and handled async so you really are only serialized by the write of responses. Do you have a recommendation on the buffer size we should have for the `responseChan`

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+// Handle registers the handler for the giveen message id
+func (mux *Mux) Handle(id prot.MessageIdentifier, handler Handler) {
+	mux.mu.Lock()
+	defer mux.mu.Unlock()
+
+	if handler == nil {
+		panic("bridge: nil handler")
+	}
+
+	if mux.m == nil {
```
jterry75: It seems like the go pattern is to lazy initialize this. I guess I don't feel strongly do you?

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+		mux.m = make(map[prot.MessageIdentifier]Handler)
 	}
+
+	if _, ok := mux.m[id]; ok {
+		logrus.Infof("bridge: overwriting bridge handler for type: %d", id)
```
jterry75: Done.

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+	return new(Mux)
 }
 
-// NewBridge produces a new bridge struct using the given Transport and Core
-// interfaces.
-func NewBridge(tport transport.Transport, coreint core.Core) *bridge {
-	return &bridge{
-		tport:   tport,
-		coreint: coreint,
+// Handle registers the handler for the giveen message id
```
jterry75: Done.

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
-	for {
-		// Messages from the HCS come as a header, which contains information
-		// about the message (including its size), and a message body.
-		// The header and body contain the operation to be performed, as well
-		// as any information needed to perform the operation.
-		message, header, err := readMessage(b.commandConn)
-		if err != nil {
-			return errors.Wrap(err, "failed to read a message from the HCS")
-		}
+	if mux.m == nil {
```
jterry75: In the *http* package they call this the *DefaultHandler*. I suppose we dont need one but it does make it easier to catch if the host supports something that we don't have implemented in the GCS yet we will get a bridge response automatically. If we don't have such a case we have to panic or something.

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+
+// Request is the bridge request that has been sent
+type Request struct {
+	Header  *prot.MessageHeader
+	Message []byte
+}
+
+// ResponseWriter is the dispatcher used to construct the Bridge response
+type ResponseWriter interface {
+	// Header is the request header that was requested
```
jterry75: The go linter threw me off here. Your not supposed to use punctuation in errors but in comments it doesn't care. I added them.

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+	Header() *prot.MessageHeader
+	// Write a successful response message.
+	Write(interface{})
+	// Error writes the provide error as a response to the message.
```
jterry75: Done.

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+// ways analogous to go's `http` package and multiplexer.
+//
+// It has two fundamentally different dispatch options:
+//
+// 1. Request/Response where by using the `Handler` a request
+//    of a given type will be dispatched to the apprpriate handler
+//    and an appropriate `ResponseWriter` will respond to exactly
+//    that request that caused the dispatch.
+//
+// 2. `PublishNotification` where by a notification that was not initiated
```
jterry75: Removed.

service/gcs/bridge/bridge.go
```diff
@@ -14,385 +16,563 @@ import (
...
+	Transport transport.Transport
+
+	// Handler to invoke when messages are received
+	Handler Handler
+
+	// commandConn is the Connection the bridge receives commands (such as
+	// ComputeSystemCreate) over.
+	commandConn transport.Connection
+
+	responseChan chan bridgeResponse
```
jterry75: Done.

service/gcs/bridge/bridge.go
jterry75: Ha! That was a bug for sure. I suppose if I unit tested the read failures I would have caught that. :) 

I think continue is the right call so that we can decide in the error chan handling what to do rather than in the request chan which doesn't know the full picture.

service/gcs/bridge/bridge.go
jterry75: It appears that you are at least warned not to do this given its a range variable. `range variable req captured by func literal` if you dont pass it via a variable.

service/gcs/bridge/bridge.go
jterry75: Made the comment more clear.

service/gcs/bridge/bridge_unit_test.go
jterry75: ginkgo doesn't seem particularly suited for unit testing because each test assumes a very specific set of handlers/errors/fakes/etc. What ends up happening is you have to put a ton of global state in the context's that some tests use and others don't. It seems better to not use ginkgo for the unit tests parts. Do you have a strong objection here?

service/gcs/bridge/bridge_unit_test.go
jterry75: Removed.

service/gcs/bridge/bridge_unit_test.go
jterry75: Done.

service/gcs/core/gcs/gcs.go
jterry75: Done.

service/gcs/core/gcs/gcs_test.go
jterry75: Done.

service/gcs/core/gcs/gcs_test.go
jterry75: Done.

service/gcs/core/gcs/gcs_test.go
jterry75: Done.

service/gcs/core/mockcore/mockcore.go
jterry75: Agreed.

service/gcs/bridge/bridge.go
beweedon: Hmm, OK. I guess I'm less interesting in the bridge/core distinction and more in separating out JSON unmarshaling from the core logic which takes structs without having to worry about how they were produced. So I'm cool with unmarshaling not being in the "bridge", as long as the GCS core methods don't have to perform the unmarshaling themselves (the unmarshaling is done either by the handler or helper functions or whatever). It seems like we're on the same page I think, I was just considering the handlers to be part of the "bridge". :)

service/gcs/bridge/bridge.go
beweedon: Hooray!

service/gcs/core/gcs/gcs.go
beweedon: I'm fine with that. If we want to have more detailed information we can add it pretty easily later.

service/gcs/bridge/bridge.go
beweedon: I don't feel super strongly, but it does seem better to me to guarantee the map is initialized on construction. That way we don't need any nil checks. The map is pretty much always going to be initialized anyway, and creating an empty map sooner shouldn't really be a performance problem.

service/gcs/bridge/bridge.go
beweedon: Ah, I see. It would be `nil` if no handlers have been registered yet. I think this is related to just initializing the map when `mux` is created. That removes the need for any special casing like this.

service/gcs/bridge/bridge.go
beweedon: Oh, weird.

service/gcs/bridge/bridge.go
beweedon: Oh, OK! It makes sense that would be a problem for range variables! Good thing Go catches that!

service/gcs/bridge/bridge_unit_test.go
beweedon: I probably need to look at your tests more closely, but I feel like it should be fine to set handlers/errors/etc. which are more specific to individual tests deeper in the closure stack where those tests are. Like, put state that's only used by certain tests only in the scope for those tests. Ginkgo is very much intended for unit testing :). It also gets rid of a lot of the code duplication going on in the "testing" tests. I think in general we should try to use the same testing framework for the whole GCS if we can.

jterry75:
@beweedon - Not sure where we are on this. Are you waiting on me?

beweedon:
Sorry, it's super difficult to keep track of comments on Github! My two outstanding pieces of feedback were not initializing the map when Mux is initialized, and not writing the unit tests using Ginkgo. My comments on both are above. If you want to talk about them offline or start a new comment thread about them, feel free, because it's really easy for stuff to get buried here otherwise :)

jterry75:
Fixed merge conflict.

beweedon:
Are you going to remove the nil checks for `mux.m`? The one that returns `NotSupportedHandler` is definitely not needed anymore if the map should always be initialized, and the panic when `mux.m` is nil seems a bit extraneous. These are my only remaining comments, though, and otherwise it looks good :)

jterry75:
@beweedon - I thought I already did that. Just submitted the fix.

beweedon:
Perfect! Thanks for adding a test!

jterry75:
@soccerGB - any comments? I would like to get this in so I can submit the rest of the changes here. Thanks

Issue #114, opened 2017-08-16
nginx is not working due to missing /dev/stdout and /dev/stderr


rn:
Using master of OpenGCS (`2dad06f038530dc305fb139a281507f1002e7e51`) and latest master from docker (see below) running `nginx` fails:
```
PS> ./docker run nginx
nginx: [alert] could not open error log file: open() "/var/log/nginx/error.log" failed (6: No such device or address)
2017/08/16 16:59:18 [emerg] 1#1: open() "/var/log/nginx/error.log" failed (6: No such device or address)
```
I noticed this on earlier versions as well.

The `nginx` container has symlinks to `/dev/stdout` and `/dev/stderr`:
```
# ls -l var/log/nginx/
total 0
lrwxrwxrwx 1 root root 11 Jul 26 07:34 access.log -> /dev/stdout
lrwxrwxrwx 1 root root 11 Jul 26 07:34 error.log -> /dev/stderr
```
and when run as `./docker run -ti nginx` on LCOW it works.

`docker run nginx` on Linux works and doing a `runc exec` in to the `nginx` container on Linux (well Docker for Mac actually) we see:
```
runc exec -t 86eb48a37c971e3912879f8415fb6362a5cdda8d0dd0fb16ecc3e9e2dc4ba070 ls -l /dev
total 0
lrwxrwxrwx 1 root root   11 Aug 16 17:26 core -> /proc/kcore
lrwxrwxrwx 1 root root   13 Aug 16 17:26 fd -> /proc/self/fd
crw-rw-rw- 1 root root 1, 7 Aug 16 17:26 full
drwxrwxrwt 2 root root   40 Aug 16 17:26 mqueue
crw-rw-rw- 1 root root 1, 3 Aug 16 17:26 null
lrwxrwxrwx 1 root root    8 Aug 16 17:26 ptmx -> pts/ptmx
drwxr-xr-x 2 root root    0 Aug 16 17:26 pts
crw-rw-rw- 1 root root 1, 8 Aug 16 17:26 random
drwxrwxrwt 2 root root   40 Aug 16 17:26 shm
lrwxrwxrwx 1 root root   15 Aug 16 17:26 stderr -> /proc/self/fd/2
lrwxrwxrwx 1 root root   15 Aug 16 17:26 stdin -> /proc/self/fd/0
lrwxrwxrwx 1 root root   15 Aug 16 17:26 stdout -> /proc/self/fd/1
[...]

PR #118, opened 2017-08-23
Implement stdio pipe relay for container processes

jterry75:
Resolves: #114

### DIFF
--- a/service/gcs/stdio/stdio.go
+++ b/service/gcs/stdio/stdio.go
@@ -6,8 +6,8 @@ import (
 	"sync"
 
 	"github.com/Microsoft/opengcs/service/gcs/transport"
-	"github.com/sirupsen/logrus"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 // ConnectionSet is a structure defining the readers and writers the Core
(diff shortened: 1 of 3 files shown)


jterry75:
I have verified through docker that this works as expected. A call to `readlink -f /proc/self/fd/{0, 1, 2}` responds with the expected `pipe:[inode]`. This matches linux behavior and unblocks nginx and others who failed with a `vsock:[inode]` descriptor.

service/gcs/stdio/stdio.go
```diff
@@ -99,6 +99,136 @@ func (s *ConnectionSet) Files() (_ *FileSet, err error) {
...
+		pr.pipes[4], pr.pipes[5], err = os.Pipe()
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to create stderr pipe relay")
+		}
+	}
+	return pr, nil
+}
+
+// PipeRelay is a relay built to expose a pipe interface
+// for stdin, stdout, stderr ontop of a ConnectionSet.
```
beweedon: nit: "on top"

service/gcs/stdio/stdio.go
```diff
@@ -99,6 +99,136 @@ func (s *ConnectionSet) Files() (_ *FileSet, err error) {
...
+type PipeRelay struct {
+	wg sync.WaitGroup
+	s  *ConnectionSet
+	// pipes format is stdin [0 read, 1 write], stdout [2 read, 3 write], stderr [4 read, 5 write].
+	pipes [6]*os.File
```
beweedon: It would be clearer to me to maybe have three sets of pipes (one for stdin, one for stdout, one for stderr). You can represent each pipe as an array or a struct or whatever works, but I just think separating them out makes the code clearer and reduces the risk of errors putting the wrong index somewhere.

service/gcs/stdio/stdio.go
```diff
@@ -99,6 +99,136 @@ func (s *ConnectionSet) Files() (_ *FileSet, err error) {
...
+// Start starts the relay operation. The caller must call Wait to wait
+// for the relay to finish and release the associated resources.
+func (pr *PipeRelay) Start() {
+	if pr.s.In != nil {
+		pr.wg.Add(1)
+		go func() {
+			if _, err := io.Copy(pr.pipes[1], pr.s.In); err != nil {
+				logrus.Errorf("error copying stdin to pipe: %s", err)
+			}
+			if err := pr.pipes[1].Close(); err != nil {
```
beweedon: I'm probably missing something, just not sure why you don't close both the src and dest to io.Copy after it exits.

service/gcs/stdio/stdio.go
```diff
@@ -99,6 +99,136 @@ func (s *ConnectionSet) Files() (_ *FileSet, err error) {
...
+
+// Wait waits for the relaying to finish and closes the associated
+// pipes and connections.
+func (pr *PipeRelay) Wait() {
+	// Close stdin so that the copying goroutine is safely unblocked; this is necessary
+	// because the host expects stdin to be closed before it will report process
+	// exit back to the client, and the client expects the process notification before
+	// it will close its side of stdin (which io.Copy is waiting on in the copying goroutine).
+	if pr.s.In != nil {
+		if err := pr.s.In.CloseRead(); err != nil {
```
beweedon: Do we really want to call CloseRead as soon as Wait is called? This makes Wait seem like it's partially acting like Close.

jstarks: We need to do something to unblock the copy goroutine after the process exits. That's what the `CloseRead` is for... Maybe there's a better name than `Wait`.

beweedon: Maybe EndStdioAndWait or something? It could also make sense to pull the CloseRead code into either a different function, or just stick it in *process.Wait. I'm cool with any option, I think just calling it Wait implies it doesn't perform any close operations.

jstarks: I don't see a reason to separate this functionality. Maybe Justin can come up with a better name that is still Go-ish.

service/gcs/stdio/stdio.go
jterry75: Done.

service/gcs/stdio/stdio.go
jterry75: This is completely internal so its pretty safe to use this way. I would prefer to leave it.

service/gcs/stdio/stdio.go
jterry75: The EOF needs to be set to the child process. So closing the `write` side does this. The outstanding call to `Wait()` will eventually close both sides and the sockets when out/err complete too so its fine here.

service/gcs/stdio/stdio.go
jterry75: I'm going to punt on this. I get the difference between a typical `Wait()` but I don't see a big benefit here.

jterry75:
@beweedon - I don't understand this race detection. It appears to be caused by the way ginkgo is running the `AfterEach` on a different goroutine.... Ideas?

service/gcs/stdio/stdio.go
beweedon: I mean, it would make the internal code clearer. But also I don't have strong feelings about this as long as it works :)

jterry75:
Pushed a commit to fix the race condition in the tests. This is good to go now if there is no more feedback.

PR #119, opened 2017-08-24
Tar2vhd should use -t ext4 when calling mount.

jterry75:


jhowardmsft:
LGTM

(BTW, this stops errors popping up on the serial console about failed to mount as ext2, then ext3, before finally using ext4).

PR #121, opened 2017-08-25
Update runc version to version used by docker

beweedon:
The docker update to use this version of runc can be found in the PR
https://github.com/moby/moby/pull/34356.

### DIFF
--- a/docs/customosbuildinstructions.md
+++ b/docs/customosbuildinstructions.md
@@ -13,11 +13,11 @@ A LCOW custom Linux OS image was divided into two parts: a Linux kernel module a
 
 - Use the recommended [Kconfig](../kernel/kernel_config-4.11.x) to include all LCOW necessary kernel components
 
-- Build your kernel 
+- Build your kernel
 
 
     Note:  The key delta between the upsteam default setting and above kconfig is in the area of ACPI/NIFT/NVDIMM/OverlyFS/9pFS/Vsock/HyerpV settings, which were set to be built-in instead of modules.
-           The Kconfig above is still a work in process in terms of trimming any unnecessary components from the kernel image.  
+           The Kconfig above is still a work in process in terms of trimming any unnecessary components from the kernel image.
 
 ## How to construct user-mode components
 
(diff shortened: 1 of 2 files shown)


jterry75:
Don't you need to update the vendor?

docs/customosbuildinstructions.md
```diff
@@ -13,59 +13,59 @@ A LCOW custom Linux OS image was divided into two parts: a Linux kernel module a
...
 
-              Note:this is the "runc" binary for hosting the container execution environment. 
+              Note:this is the "runc" binary for hosting the container execution environment.
               It needs to be the following release
-              runc version 1.0.0-rc3
-              commit: 992a5be178a62e026f4069f443c6164912adbf09
-              spec: 1.0.0-rc5
+              runc version 1.0.0-rc4
+              commit: 3f2f8b84a77f73d38244dd690525642a72156c64
+              spec: 1.0.0
```
soccerGB: The official commit for runc v1.0.0-rc4 is https://github.com/opencontainers/runc/commit/2e7cfe036e2c6dc51ccca6eb7fa3ee6b63976dcd
It might be a good idea to use it instead of 3f3f... or the one I added in setup_test_env ( 882d8eaba6682179e397c78d5a40e45fa4d571ce)

Could you syn them up?

soccerGB: RunC is a separate binary that we pulled into linux initrd. This binary is not part of the opengcs repo.
So, no need to change any vendor file.

beweedon: Do you know what the advantage of using the rc4 commit rather than the commit docker uses would be?

beweedon:
(also in reply to Cheng-mean) The vendored runc is used by the docker/containerd packages. I didn't change the vendored version because I mostly wanted this to be a testing/documentation change, but you're right, I should probably change the vendored version as well for consistency/rather than making another change.

beweedon:
Actually, on second thought, because the vendored runc package is only used by docker and containerd, if we update runc we should probably also update those. So I think this is better in another change just for generically updating the vendored dependencies which need to be updated. I'll make an issue.

beweedon:
#122

docs/customosbuildinstructions.md
soccerGB: The official RunC release one should be more stable.

soccerGB:
It looks like the output of "runc -v" has changed 
from
# ./runc -v
runc version 1.0.0-rc3
commit: 2d41c047c83e09a6d61d464906feb2a2f3c52aa4
spec: 1.0.0-rc5

to 
# ./runc -v
runc version spec: 1.0.0

it would be great if our doc could reflect that to avoid confusion.

beweedon:
Ah, OK, I see that the doc was meant to represent the output of "runc -v" now. So, actually, runc installed with Go has always had the one-line output as far as I can remember. If you build it using the makefile in the runc repo, however, it produces that extra output information (I assume because it's building a "developer" build or something). I can update customosbuildinstructions.md to reflect the two possible outputs.

docs/customosbuildinstructions.md
beweedon: Talked offline, going to use the version docker uses at least for now, since all of our testing, etc. is on docker master.

PR #125, opened 2017-08-31
Implements the bridge handler tests

jterry75:
1. Updates the bridge_test suite to test just the handler code and all error/success cases.
2. Cleans up a little bit of the bridge handler code.


service/gcs/bridge/bridge_handler_test.go
```diff
@@ -0,0 +1,800 @@
...
+		coreint: &mockcore.MockCore{
+			Behavior: mockcore.Error,
+		},
+	}
+	tb.createContainer(rw, req)
+
+	verifyResponseError(t, rw)
+}
+
+func Test_CreateContainer_Success_WaitContainer_Failure(t *testing.T) {
```
beweedon: Isn't this a success test?

service/gcs/bridge/bridge_handler_test.go
```diff
@@ -0,0 +1,800 @@
...
+
+	tb := &Bridge{
+		coreint: &mockcore.MockCore{
+			Behavior: mockcore.Success,
+		},
+	}
+	tb.modifySettings(rw, req)
+
+	verifyResponseSuccess(t, rw)
+}
```
beweedon: I think my main suggestion for the tests is that they test that the correct values were unpacked into mockcore inconsistently. I think it would be a good idea to add checks that the values were properly unmarshaled into the success tests (e.g. `Test_ModifySettings_CoreSucceeds_Success`), so that we have validation that the correct values make it from the message to the Core.

service/gcs/bridge/bridge_handler_test.go
jterry75: This name is to imply the CreateContainer succeeds but the WaitContainer notification fails.

beweedon: I might be missing something, but where in the test does it test for failure?

service/gcs/bridge/bridge_handler_test.go
jterry75: For the success cases I will check all of the params. Working on the fix.

beweedon: Also, sorry, *consistently :)

beweedon: And thanks! :)

jterry75:
Fixed the test race condition.

service/gcs/bridge/bridge_handler_test.go
beweedon: Does the comment you added describe how the test is testing for failure? I'm not sure I understand exactly what it's saying, and just want to make sure.

service/gcs/bridge/bridge_handler_test.go
```diff
@@ -0,0 +1,791 @@
...
+			ProcessParameters: "",
+		},
+	}
+
+	req, rw := setupRequestResponse(t, prot.ComputeSystemExecuteProcessV1, r)
+
+	tb := new(Bridge)
+	tb.execProcess(rw, req)
+
+	verifyResponseJSONError(t, rw)
```
beweedon: You should probably test for correct ActivityId on both success and failure as well. I think not filling in ActivityId for failures is the cause on the invalid JSON error we've been seeing.

jterry75: Fixed in latest push.

jterry75:
@gupta-ak - since @soccerGB is out mind reviewing and merging?

PR #128, opened 2017-09-01
gcs: Mount plan9 shares with fd transport

jstarks:
This change eliminates the need for the plan9 vsock transport by
connecting to the plan9 server in GCS and using the plan9 fd
transport.

### DIFF
--- a/service/gcs/core/gcs/storage.go
+++ b/service/gcs/core/gcs/storage.go
@@ -253,25 +253,34 @@ func (c *gcsCore) unmountMappedVirtualDisks(disks []prot.MappedVirtualDisk) erro
 	return nil
 }
 
-// mountMappedDirectories mounts the given mapped directories using a Plan9
+// mountMappedDirectory mounts the given mapped directory using a Plan9
 // filesystem with the given options.
-func (c *gcsCore) mountMappedDirectories(dirs []prot.MappedDirectory) error {
-	for _, dir := range dirs {
-		if !dir.CreateInUtilityVM {
-			return errors.New("we do not currently support mapping directories inside the container namespace")
-		}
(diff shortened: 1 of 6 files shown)


jstarks:
I've validated that this works. I suggest we use this approach and stop trying to push the new vsock v9fs transport to the upstream kernel.

service/gcs/core/gcs/storage.go
```diff
@@ -253,25 +253,34 @@ func (c *gcsCore) unmountMappedVirtualDisks(disks []prot.MappedVirtualDisk) erro
...
-			return errors.Wrapf(err, "failed to mount directory for mapped directory %s", dir.ContainerPath)
-		}
+func (c *gcsCore) mountMappedDirectory(dir *prot.MappedDirectory) error {
+	if !dir.CreateInUtilityVM {
+		return errors.New("we do not currently support mapping directories inside the container namespace")
+	}
+	if err := c.OS.MkdirAll(dir.ContainerPath, 0700); err != nil {
+		return errors.Wrapf(err, "failed to create directory for mapped directory %s", dir.ContainerPath)
+	}
+	conn, err := c.vsock.Dial(dir.Port)
```
soccerGB: This change eliminates the need for the 9pfs patch, which is good.

The only question I have is: are there any cleanup need to be done in unmountMappedVirtualDisks when this mapping is no longer needed?

soccerGB: Other than that, the change looks good to me.

jstarks: No, there's no other cleanup necessary. What happens is that the kernel references the file descriptor as part of the mount call, so after the mount call returns we can close the fds immediately. Then after unmount the kernel will release its reference and the socket will close without any additional code in GCS.

jterry75:
LGTM

Issue #131, opened 2017-09-07
GCS will delete contents of mapped directories on exit if they are under /tmp/gcs/<containerID>/


beweedon:
This is because the GCS apparently doesn't unmount mapped directories during cleanup (see https://github.com/Microsoft/opengcs/blob/master/service/gcs/core/gcs/cleanup.go). We should add code to perform this unmounting during the cleanup phase.

@jterry75 @jhowardmsft

jstarks:
Holy smokes! As an extra precaution, can we avoid crossing file-system boundaries when we delete /tmp/gcs/.../ ?

simonferquel:
@jstarks I agree. If for whatever reason the unmounts fails, we need to be sure the content does not get erased by some code further down the path. We have been hurt very hard by a similar issue at some point with Docker for Windows, and really, you don't want to have to patch this kind of bug while in production.

jterry75:
Do you have a proposed mitigation? I agree there is a risk here for sure when unmount fails but I don't know how I can cleanup the container in this case. Any deletes to the rootfs will also delete the mounted folder

simonferquel:
with unix fs semantics, I would assume that `mv <mountpoint> /to-not-remove` might work... In the case of docker for Windows we decided that if an unmount fails, we don't remove the mount points at all.

gupta-ak:
Reopening for further discussion.

After unmounting, you might be able to do a rename(rootfs, rootfs-removing) or something to make sure you're not crossing file system boundaries and then delete it.

jhowardmsft:
Would it just be easier to put it back at /binds/containerID/... instead? The location is under the control of docker, so it's a simple fix. At least that would guarantee not to be cleaned by GCS.

gupta-ak:
I think the issue is that the gcs does the equivalent of `rm -rf rootfs` which will delete everything including layer + plan9 data if the previous unmounts did not succeed. It doesn't matter where the mounts sources are located, because they eventually have to be mounted inside `rootfs` so that the container can access them. 

 Right now the current cleanup code tries to unmount the mapped disks, the mapped directories and the layers and then does `rm -rf rootfs`. If any of the unmounts fails, then it just logs it and continues forward, which can corrupt data on the delete.

It makes sense to me to have the same behavior as Docker for Windows where we simply don't clean up if the unmounts fail. Otherwise, we could corrupt the layers.

rn:
[...]

PR #132, opened 2017-09-07
Fixes an issue on cleanup with mapped directories

jterry75:
Resolves: #131

### DIFF
--- a/service/gcs/core/gcs/cleanup.go
+++ b/service/gcs/core/gcs/cleanup.go
@@ -30,14 +30,23 @@ func (c *gcsCore) cleanupContainer(containerEntry *containerCacheEntry) error {
 			errToReturn = err
 		}
 	}
-
+	directoryMap := containerEntry.MappedDirectories
+	directories := make([]prot.MappedDirectory, 0, len(directoryMap))
+	for _, directory := range directoryMap {
+		directories = append(directories, directory)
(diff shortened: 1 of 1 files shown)


jterry75:
@jhowardmsft - Has verified the fix

jhowardmsft:
LGTM

service/gcs/core/gcs/cleanup.go
```diff
@@ -20,7 +20,7 @@ func (c *gcsCore) cleanupContainer(containerEntry *containerCacheEntry) error {
 	}
 
 	diskMap := containerEntry.MappedVirtualDisks
-	disks := make([]prot.MappedVirtualDisk, 0, len(diskMap))
```
beweedon: Correct me if I'm wrong, but if we use make like this, doesn't it create a slice of length `len(diskMap)` full of default elements, and then `append` just adds on to this slice? So we'd end up with a slice half full of default `MappedVirtualDisk`s?

beweedon: You could also do it by not specifying the capacity as your have, and assigning the disk/dir at the correct index in the assignment loop as well. I honestly like that better I think :)

jterry75: Oh good call.

PR #133, opened 2017-09-08
Add a retry when connecting stdin/stdout/stderr

gupta-ak:
Temp fix until underlying vsock issue is solved.

### DIFF
--- a/service/gcs/transport/vsock.go
+++ b/service/gcs/transport/vsock.go
@@ -1,9 +1,11 @@
 package transport
 
 import (
-	"github.com/sirupsen/logrus"
+	"fmt"
+	"time"
+
 	"github.com/linuxkit/virtsock/pkg/vsock"
-	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


jterry75:
I think you need to add this when calling Dial for the p9fs mapped directories too

gupta-ak:
Are you talking about this: https://github.com/Microsoft/opengcs/blob/ab442e9a042246e5e6c87102ed56faa6d3925458/service/gcs/core/gcs/storage.go#L265

despite the name, i think it's a transport object:
https://github.com/Microsoft/opengcs/blob/ab442e9a042246e5e6c87102ed56faa6d3925458/service/gcs/core/gcs/gcs.go#L36

jterry75:
`transport.Transport` is the interface. If you look in main.go you see that its initialized with a `VsockTransport{}` so it could have the same issue on `Dial`.

gupta-ak:
isn't that what I just changed?
`func (t *VsockTransport) Dial(port uint32) (Connection, error) { `

Is there another implementation of the Dial interface that I'm missing?

gupta-ak:
oh it also turns out that the virtsock library masked the ETIMEDOUT, so now I just retry on any error.

PR #137, opened 2017-09-14
Fixes potential delete on unmounts failure

jterry75:
Will only destroy the container storage in the UVM if all mounts are successfully unmounted. Without this we have no way of knowing if it safe to delete the files or if this could have an affect on the host files.

Resolves: #131

@jhowardmsft - FYI

### DIFF
--- a/service/gcs/core/gcs/cleanup.go
+++ b/service/gcs/core/gcs/cleanup.go
@@ -47,11 +47,18 @@ func (c *gcsCore) cleanupContainer(containerEntry *containerCacheEntry) error {
 			errToReturn = err
 		}
 	}
-	if err := c.destroyContainerStorage(containerEntry.ID); err != nil {
-		logrus.Warn(err)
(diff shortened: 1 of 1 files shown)


service/gcs/core/gcs/cleanup.go
```diff
@@ -47,11 +47,18 @@ func (c *gcsCore) cleanupContainer(containerEntry *containerCacheEntry) error {
 			errToReturn = err
 		}
 	}
-	if err := c.destroyContainerStorage(containerEntry.ID); err != nil {
-		logrus.Warn(err)
-		if errToReturn == nil {
-			errToReturn = err
+
+	// We only do cleanup
```
beweedon: Is this supposed to say something like "We only do cleanup if unmounting succeeds"?

jterry75: Lol. Guess I got sidetracked. Fixed.

Issue #130, opened 2017-09-07
Move runc logs to per-container directory


beweedon:
Currently, we have global runc logs for the entire utility VM at `/var/run/gcsrunc/log.log`. We should probably move logs to `/tmp/gcs/<containerID>/runc.log`, so it will be in the same directory as `gcs.log`. This should be a pretty simple change to the runc wrapper code.

jhowardmsft:
Yes, yes, yes!

PR #139, opened 2017-09-15
Moves runc logs to per container id

jterry75:
Resolves: #130

### DIFF
--- a/client/process.go
+++ b/client/process.go
@@ -136,7 +136,8 @@ func (config *Config) DebugGCS() {
 		cmd += debugCommand("ls -l /tmp/gcs/*")
 		cmd += debugCommand("cat /tmp/gcs/*/config.json")
 		cmd += debugCommand("ls -lR /var/run/gcsrunc")
-		cmd += debugCommand("cat /var/run/gcsrunc/log.log")
+		cmd += debugCommand("cat /tmp/gcs/global-runc.log")
+		cmd += debugCommand("cat /tmp/gcs/*/runc.log")
 		cmd += debugCommand("ps -ef")
 		cmd += `"`
 	}
--- a/service/gcs/runtime/runc/utils.go
+++ b/service/gcs/runtime/runc/utils.go
@@ -61,9 +61,23 @@ func (r *runcRuntime) makeContainerDir(id string) error {
 	return nil
 }
 
+// getLogDir gets the path to the runc logs directory.
+func (r *runcRuntime) getLogDir(id string) string {
+	return filepath.Join(r.runcLogBasePath, id)
+}
+
+// makeLogDir creates the runc logs directory if it doesnt exist.
+func (r *runcRuntime) makeLogDir(id string) error {
+	dir := r.getLogDir(id)
+	if err := os.MkdirAll(dir, os.ModeDir); err != nil {
+		return errors.Wrapf(err, "failed making runc log directory for container %s", id)
+	}
+	return nil
+}
+
 // getLogPath returns the path to the log file used by the runC wrapper.
-func (r *runcRuntime) getLogPath() string {
-	return filepath.Join(containerFilesDir, "log.log")
+func (r *runcRuntime) getLogPath(id string) string {
+	return filepath.Join(r.getLogDir(id), "runc.log")
 }
 
 // processExists returns true if the given process exists in /proc, false if
--- a/service/gcs/runtime/runc/utils_test.go
+++ b/service/gcs/runtime/runc/utils_test.go
@@ -17,7 +17,7 @@ var _ = Describe("Utils", func() {
 	)
 
 	BeforeEach(func() {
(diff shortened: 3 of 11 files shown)


service/gcs/runtime/runc/utils_test.go
```diff
@@ -147,11 +147,12 @@ var _ = Describe("Utils", func() {
 			expectedPath string
 			actualPath   string
 		)
+		id := "atestid"
 		BeforeEach(func() {
-			expectedPath = "/var/run/gcsrunc/log.log"
+			expectedPath = "/tmp/gcs/" + id + "runc.log"
```
beweedon: You'll probably notice this when the tests are done, but I think there should be a slash before runc.log here.

service/gcs/runtime/runc/utils.go
```diff
@@ -62,8 +62,8 @@ func (r *runcRuntime) makeContainerDir(id string) error {
 }
 
 // getLogPath returns the path to the log file used by the runC wrapper.
-func (r *runcRuntime) getLogPath() string {
-	return filepath.Join(containerFilesDir, "log.log")
+func (r *runcRuntime) getLogPath(id string) string {
+	return filepath.Join("/tmp/gcs", id, "runc.log")
```
beweedon: nit: It would be nice to have "/tmp/gcs" in some constant somewhere, either by having an actual constant or using `getStorageRootPath()`

service/gcs/runtime/runc/utils_test.go
jterry75: Ha. Yea your right. I wish I could run these locally

rn:
doesn't this require changes to https://github.com/Microsoft/opengcs/blob/master/client/process.go#L139 as well?

jterry75:
@rn - good point
@beweedon - I'm not sure this change makes sense given the debugging work. We either need to `T` the logs to a global and a per container log or just leave them as global (ie: revert this and close the issue). Otherwise we wont be able to get the full picture from a debug call. You would need to call it once per container id.

beweedon:
My reasoning for this change was that it would be good to have runc logs separated out by container because when debugging a failure, you'd probably want to localize your search to individual containers at a time rather than every piece of logging done by runc for all containers. This is less relevant now, and more useful when we support multiple containers per uvm. Also, this change allows all logs to be coalesced a bit under `/tmp`, rather than spread out to `/var/run/gcsrunc`.

@jhowardmsft, I know you were also enthusiastic about this change :). You know a lot more about this from the docker end than I do, so I'm not sure what your thoughts are.

jhowardmsft:
If I'm understanding the comments above correctly, it should be a straightforward change to the debug code to just cat the right log. Remember the daemon only runs a single container in a utility VM (even though we support >1 through HCS).

jterry75:
@jhowardmsft - The issue is that we don't actually know the container id in the debug code. If we move it from a global location we could certainly `cat` them all (I get that it is just 1 for now) but I don't know how you would specify the actual id your interested in.

jhowardmsft:
I'm already one step ahead of you :) Spoke in person.....

jterry75:
Added an extra call to make the container dir for the tests but in most cases this won't do anything. Now the tests are passing and the PR should be good for final review.

client/process.go
```diff
@@ -136,7 +136,8 @@ func (config *Config) DebugGCS() {
 		cmd += debugCommand("ls -l /tmp/gcs/*")
 		cmd += debugCommand("cat /tmp/gcs/*/config.json")
 		cmd += debugCommand("ls -lR /var/run/gcsrunc")
```
jhowardmsft: This should probably change too now.... (I don't think /var/run/gcs* is used anymore)

jterry75: It still is used. We write the pid file there for create container/process

beweedon: It isn't used for logs anymore, though.

jterry75: I agree but isn't it usefully to get an output of all the pid files? IE: all the running process in runc?

jhowardmsft: Sure. The more the merrier :)

jterry75: Alternatively I can move the use of this dir to be /tmp/gcs/(id)/ too?

beweedon: If we don't care about the runc wrapper being a generic wrapper that can be used outside of this project, I think it makes sense to merge it as much as possible with the GCS. Also, I guess in addition to pid files there are also process.json files under /var/run/gcsrunc, which could definitely be useful for logging.

jterry75: @beweedon - Want to open an issue for merging the two then? I don't see any use cases for a runc wrapper library. Its primarily just for us and how we use runc

beweedon: Sure thing!

beweedon: #144

jterry75:
@soccerGB - anything left for me to do here?

PR #152, opened 2017-10-31
Adds stack dump support

jterry75:


jterry75:
@jhowardmsft - Pulled this out so we can merge it while I continue to investigate.

jterry75:
@soccerGB - Can we merge?

jterry75:
@jhowardmsft - Are you ok if I merge this and then yours? We have verified it works and @soccerGB seems to be out.

Issue #153, opened 2017-10-31
GCS not always returning exit code for a process


jterry75:
@jhowardmsft - Logging for ya.

Running a command like:
docker run --rm ubuntu bash exit 123

Will return in an exit code from the HCS of 0.

jhowardmsft:
Ha, yeah I forgot to open. Thanks!

Issue #126, opened 2017-08-31
WaitProcess should honor TimeoutInMS


jterry75:
We currently don't actually honor the timeout on a wait process call and it will wait forever. If the users passes a timeout we should cancel the wait after that amount of time.

PR #157, opened 2017-11-07
Fixing issues with process exit

jterry75:
Resolves: #153 - Init process exit codes are not always returning because the container exited notification returns before the WaitProcess call comes in causing the HCS to teardown the container.
Resolves: #126 - WaitProcess should honor the timeout associated with the call and return the proper error response on timeout.

### DIFF
--- a/service/gcs/bridge/bridge.go
+++ b/service/gcs/bridge/bridge.go
@@ -347,7 +347,7 @@ func (b *Bridge) createContainer(w ResponseWriter, r *Request) {
 
 	exitCodeFn, err := b.coreint.WaitContainer(id)
 	if err != nil {
-		w.Error(request.ActivityID, err)
+		logrus.Error(err)
 	}
 
 	go func() {
--- a/service/gcs/bridge/bridge.go
+++ b/service/gcs/bridge/bridge.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/Microsoft/opengcs/service/gcs/core"
 	"github.com/Microsoft/opengcs/service/gcs/gcserr"
@@ -244,7 +245,8 @@ func (b *Bridge) ListenAndServe() (conerr error) {
 				requestErrChan <- errors.Wrap(err, "bridge: failed reading message payload")
 				continue
 			}
-			logrus.Infof("bridge: read message '%s'\n", message)
+			logrus.Infof("bridge: read message type: %v", header.Type)
+			logrus.Infof("bridge: read message '%s'", message)
 			requestChan <- &Request{header, message}
 		}
 	}()
@@ -342,14 +344,14 @@ func (b *Bridge) createContainer(w ResponseWriter, r *Request) {
 		},
 		SelectedProtocolVersion: prot.PvV3,
 	}
-	w.Write(response)
+
+	exitCodeFn, err := b.coreint.WaitContainer(id)
+	if err != nil {
+		w.Error(request.ActivityID, err)
+	}
 
 	go func() {
-		exitCode, err := b.coreint.WaitContainer(id)
-		if err != nil {
-			logrus.Error(err)
-			return
-		}
+		exitCode := exitCodeFn()
 		notification := &prot.ContainerNotification{
 			MessageBase: &prot.MessageBase{
 				ContainerID: id,
@@ -362,6 +364,8 @@ func (b *Bridge) createContainer(w ResponseWriter, r *Request) {
 		}
 		b.PublishNotification(notification)
 	}()
+
+	w.Write(response)
 }
 
 func (b *Bridge) execProcess(w ResponseWriter, r *Request) {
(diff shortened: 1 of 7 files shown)


beweedon:
Just an FYI about the timeout waiting for the process. Currently, the HCS enforces this timeout itself, so it isn't strictly necessary to enforce it here. I'm fine with it, though, just pointing it out :)

service/gcs/core/gcs/gcs.go
```diff
@@ -117,14 +116,18 @@ func (e *containerCacheEntry) RemoveMappedDirectory(dir prot.MappedDirectory) {
...
-	ContainerID string // If "" a host process otherwise a container process.
-	exitWg      sync.WaitGroup
-	exitCode    int
-}
-
-func newProcessCacheEntry(containerID string) *processCacheEntry {
-	return &processCacheEntry{ContainerID: containerID, exitCode: -1}
+	syncRoot sync.Mutex
+	Tty      *stdio.TtyRelay
+	exitWg   sync.WaitGroup // Signaled when the process itself has exited.
```
beweedon: nit: Put comment above `exitWg` to match pattern below. Plus, I'm not sure godoc recognizes comments that aren't above the thing being commented (although I could be wrong).

> The convention is simple: to document a type, variable, constant, function, or even a package, write a regular comment directly preceding its declaration, with no intervening blank line.

service/gcs/core/gcs/gcs.go
```diff
@@ -542,32 +558,60 @@ func (c *gcsCore) ResizeConsole(pid int, height, width uint16) error {
...
+		} else {
+			entry.writersCalled = true
+		}
+		entry.syncRoot.Unlock()
+	}
+
+	f := func() int {
+		entry.exitWg.Wait()
+		if entry.isInitProcess {
+			entry.writersWg.Done()
```
beweedon: Is it safe to set this done before the write has actually been made?

service/gcs/bridge/bridge.go
```diff
@@ -362,6 +363,9 @@ func (b *Bridge) createContainer(w ResponseWriter, r *Request) {
 		}
 		b.PublishNotification(notification)
 	}()
+
+	w.Write(response)
```
beweedon: Don't we want to write the response before the go routine? That way, we don't get an exit notification before the create call returns.

service/gcs/bridge/bridge.go
```diff
@@ -362,6 +363,9 @@ func (b *Bridge) createContainer(w ResponseWriter, r *Request) {
...
+
```
beweedon: nit: Extra blank line.

service/gcs/bridge/bridge.go
```diff
@@ -244,7 +245,7 @@ func (b *Bridge) ListenAndServe() (conerr error) {
 				requestErrChan <- errors.Wrap(err, "bridge: failed reading message payload")
 				continue
 			}
-			logrus.Infof("bridge: read message '%s'\n", message)
+			logrus.Infof("bridge: read message type: '%v' message: '%s'\n", header.Type, message)
```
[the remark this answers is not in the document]
jterry75: #158 - Will fix on separate commit.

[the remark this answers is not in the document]
jterry75: That's good feedback. It is very hard to trace the additional line is easier on the eyes. Fixed

service/gcs/bridge/bridge.go
jterry75: fixed.

service/gcs/core/gcs/gcs.go
jterry75: Fixed.

service/gcs/core/gcs/gcs.go
jterry75: I agree its a race. I'm not sure how I could possibly do this. I would need to expose a "written" response channel from the bridge.Write and we would have to only release the `writersWg` when we received that message. And even that would not be *truly* enough. Did the socket actually write? Its almost like we would need a two-phase with the HCS to know. IE: Write -> HCS -> Written. Ok now release. Thoughts?

jterry75: Actually... I was wrong. The bridge serializes messages once the call to "Write" has occurred so I can do this. Let me think about how to fix it.

service/gcs/bridge/bridge.go
beweedon: Oh, right, my bad, you're right :)

service/gcs/core/gcs/gcs.go
```diff
@@ -579,40 +587,77 @@ func (c *gcsCore) WaitContainer(id string) (func() int, error) {
...
-		// added the 1st writer at creation to ensure that the container exited
-		// notification wasn't written until at least one `WaitProcess` was called.
-		entry.syncRoot.Lock()
-		if entry.writersCalled {
-			entry.writersWg.Add(1)
-		} else {
-			entry.writersCalled = true
-		}
-		entry.syncRoot.Unlock()
+		entry.writersSyncRoot.Lock()
```
beweedon: Are these locks necessary?

service/gcs/core/gcs/gcs.go
```diff
@@ -579,40 +587,77 @@ func (c *gcsCore) WaitContainer(id string) (func() int, error) {
...
+					}
+					entry.writersSyncRoot.Unlock()
+				}
+			}
+		case <-doneChan:
+			// This case means that the waiter decided to stop waiting before
+			// the process had an exit code. In this case we need to cleanup
+			// just our waiter because the no response was written.
+			if entry.isInitProcess {
+				entry.writersSyncRoot.Lock()
```
beweedon: Same about the locks here.

service/gcs/core/gcs/gcs.go
jterry75: Yes. It is not safe in golang to use a `WaitGroup` where you have someone who has called `Wait()` and others who might call `Add/Done`

service/gcs/core/gcs/gcs.go
jterry75: Same.

service/gcs/core/gcs/gcs.go
beweedon: Ah, interesting. Good to know!

service/gcs/core/gcs/gcs.go
```diff
@@ -579,40 +587,77 @@ func (c *gcsCore) WaitContainer(id string) (func() int, error) {
...
-		// notification wasn't written until at least one `WaitProcess` was called.
-		entry.syncRoot.Lock()
-		if entry.writersCalled {
-			entry.writersWg.Add(1)
-		} else {
-			entry.writersCalled = true
-		}
-		entry.syncRoot.Unlock()
+		entry.writersSyncRoot.Lock()
+		entry.writersWg.Add(1)
```
beweedon: I think this is my last question, but should this `Add` be conditioned on `writersCalled` being true like it was before?

jterry75: So I change the flow in order to make safer sense on timeout cases.

Every call to `WaitProcess` where it is the *init process* will increment its wait count. When it either succeeds or times out. It will indicate that its done waiting. If it indicates that its done before an exit code was given then we know it timed out and we don't do the "extra" decrement associated with the `WaitContainer` call. If it did get the exit code before the acknowledgement then we check if its the first one and do the extra decrement but not for any additional waiters so that we keep the `WaitGroup` in sync.

beweedon: Make sense! I'll approve it :)

jterry75:
Fixed the unit tests and pushed a new update.

jterry75:
@soccerGB - This is being hit by quite a few docker build scenarios with exit handling on file/folder detection. If your ok with this can we merge to master?

PR #176, opened 2017-12-19
Handle start message as noop in bridge

### LINKED ISSUE #163: Start message should be a nop rather than failing
New HCS orchestrators will send a start message to both Linux and Windows GCSs

beweedon:
The HCS will send this message to Linux VMs in the future, so we should
handle it rather than returning an error.

Resolves #163

### DIFF
--- a/service/gcs/bridge/bridge.go
+++ b/service/gcs/bridge/bridge.go
@@ -228,6 +228,9 @@ func (b *Bridge) AssignHandlers(mux *Mux, gcs core.Core) {
 	mux.HandleFunc(prot.ComputeSystemWaitForProcessV1, prot.PvV3, b.waitOnProcess)
 	mux.HandleFunc(prot.ComputeSystemResizeConsoleV1, prot.PvV3, b.resizeConsole)
 	mux.HandleFunc(prot.ComputeSystemModifySettingsV1, prot.PvV3, b.modifySettings)
+
+	// v4 specific handlers
+	mux.HandleFunc(prot.ComputeSystemStartV1, prot.PvV4, b.startContainer)
 }
 
 // ListenAndServe connects to the bridge transport, listens for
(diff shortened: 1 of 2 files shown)


service/gcs/bridge/bridge.go
```diff
@@ -368,6 +369,20 @@ func (b *Bridge) createContainer(w ResponseWriter, r *Request) {
 	w.Write(response)
 }
 
+func (b *Bridge) startContainer(w ResponseWriter, r *Request) {
+	// This is just a noop, but needs to be handled so that an error isn't
```
jterry75: Can you add a `TODO: v4 impl required` so that I remember in two weeks :)

service/gcs/bridge/bridge.go
```diff
@@ -197,6 +197,7 @@ type Bridge struct {
 func (b *Bridge) AssignHandlers(mux *Mux, gcs core.Core) {
 	b.coreint = gcs
 	mux.HandleFunc(prot.ComputeSystemCreateV1, b.createContainer)
+	mux.HandleFunc(prot.ComputeSystemStartV1, b.startContainer)
```
jterry75: Can you add a comment `v4 only`

service/gcs/bridge/bridge.go
beweedon: Wait, do we want a V4 implementation? I wasn't sure if we were doing anything in start, and just didn't want it producing an error.

jterry75: Can you point me to the docs for how you envision the new protocol working? I guess I dont understand the ordering.

soccerGB: "The HCS will send this message to Linux VMs in the future,...."
Do you mean RS4 or RS5?

jterry75:
Rebased on #180 but should only be merged after that.

beweedon:
ping

PR #178, opened 2017-12-19
Unknown messages should return E_VMCOMPUTE_UNKNOWN_MESSAGE

### LINKED ISSUE #175: Return E_VMCOMPUTE_UNKNOWN_MESSAGE on unknown message type
Right now it returns `E_NOTIMPL`.

beweedon:
They used to return E_NOTIMPL, which is different from what the Windows
GCS would return.

Resolves #175

### DIFF
--- a/service/gcs/bridge/bridge.go
+++ b/service/gcs/bridge/bridge.go
@@ -26,16 +26,16 @@ var capabilities = prot.GcsCapabilities{
 	SendInitialCreateMessage: false,
 }
 
-// NotSupported represents the default handler logic for an unmatched
-// request type sent from the bridge.
-func NotSupported(w ResponseWriter, r *Request) {
-	w.Error("", gcserr.WrapHresult(errors.Errorf("bridge: function not supported, header type: 0x%x", r.Header.Type), gcserr.HrNotImpl))
+// UnknownMessage represents the default handler logic for an unmatched request
+// type sent from the bridge.
+func UnknownMessage(w ResponseWriter, r *Request) {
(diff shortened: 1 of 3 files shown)


service/gcs/bridge/bridge.go
```diff
@@ -24,7 +24,7 @@ import (
 // NotSupported represents the default handler logic for an unmatched
 // request type sent from the bridge.
 func NotSupported(w ResponseWriter, r *Request) {
-	w.Error("", gcserr.WrapHresult(errors.Errorf("bridge: function not supported, header type: 0x%x", r.Header.Type), gcserr.HrNotImpl))
+	w.Error("", gcserr.WrapHresult(errors.Errorf("bridge: function not supported, header type: 0x%x", r.Header.Type), gcserr.HrVmcomputeUnknownMessage))
```
jterry75: I think we can only return the 'proper' code in the v4 version case right? Otherwise this is a breaking change to the HCS?

jterry75: Also, did you run the unit tests for the bridge? I would guess I had an invalid message test that would fail now.

beweedon: Yeah, that's kinda what my bold text comment in the description means :). I mostly just wanted to get this PR up and then you or I could make that necessary change after protocol negotiation is added.

beweedon: And, yeah, I updated the unit tests so they should pass.

jterry75:
Again, rebased on #180 but should not be merged before then.

beweedon:
ping

PR #180, opened 2018-01-05
 Implements version based multiplexer dispatch

jterry75:


jterry75:
Building a kernel to verify. Will update before merge

service/gcs/bridge/bridge.go
```diff
@@ -317,7 +340,55 @@ func (b *Bridge) PublishNotification(n *prot.ContainerNotification) {
 	b.responseChan <- resp
 }
 
+// negotiateProtocol was introduced in v4 so will not be called
+// with a minimum lower than that.
+func (b *Bridge) negotiateProtocol(w ResponseWriter, r *Request) {
+	if r.Version != prot.PvInvalid {
```
beweedon: I think this is handled by the routing mechanism, right? Since `negotiateProtocol` is registered with `prot.PvInvalid`, it should only be called when that's set.

service/gcs/bridge/bridge.go
```diff
@@ -317,7 +340,55 @@ func (b *Bridge) PublishNotification(n *prot.ContainerNotification) {
...
+		w.Error("", errors.Wrapf(err, "failed to unmarshal JSON in message \"%s\"", r.Message))
+		return
+	}
+
+	if request.MaximumVersion.Major < uint32(prot.PvV4) || uint32(prot.PvMax) < request.MinimumVersion.Major {
+		w.Error(request.ActivityID, gcserr.NewHresultError(gcserr.HrVmcomputeUnsupportedProtocolVersion))
+		return
+	}
+
+	min := func(x, y uint32) uint32 {
```
beweedon: Could you use `math.Min` instead?

service/gcs/bridge/bridge.go
```diff
@@ -317,7 +340,55 @@ func (b *Bridge) PublishNotification(n *prot.ContainerNotification) {
...
+		Capabilities: capabilities,
+	}
+
+	// Set our protocol selected version before return.
+	b.protVer = prot.ProtocolVersion(major)
+	w.Write(response)
+}
+
 func (b *Bridge) createContainer(w ResponseWriter, r *Request) {
+	if r.Version != prot.PvInvalid {
```
beweedon: Same thing mentioned above about the routing.

service/gcs/bridge/bridge.go
```diff
@@ -189,22 +204,30 @@ type Bridge struct {
...
-	mux.HandleFunc(prot.ComputeSystemSignalProcessV1, b.signalProcess)
-	mux.HandleFunc(prot.ComputeSystemGetPropertiesV1, b.listProcesses)
-	mux.HandleFunc(prot.ComputeSystemWaitForProcessV1, b.waitOnProcess)
-	mux.HandleFunc(prot.ComputeSystemResizeConsoleV1, b.resizeConsole)
-	mux.HandleFunc(prot.ComputeSystemModifySettingsV1, b.modifySettings)
+
+	// These are PvInvalid because they will be called previous to any protocol
+	// negotiation so they respond only when the protocols are not known.
+	mux.HandleFunc(prot.ComputeSystemNegotiateProtocolV1, prot.PvInvalid, b.negotiateProtocol)
+	mux.HandleFunc(prot.ComputeSystemCreateV1, prot.PvInvalid, b.createContainer)
```
beweedon: Currently, this means `createContainer` can't be called by a V4 HCS, because a protocol negotiation will set the version to a non-invalid value.

jterry75: Correct. After looking at it we will need a new impl for `createContainer` in v4 anyways so as a separate PR I will implement the new function.

beweedon: OK, sounds good!

service/gcs/bridge/bridge.go
jterry75: Actually the goal here is to make sure that we don't ever try to negotiate a new version. IE: if the request has updated the version once this call would error out the 2nd time. And for a v3 workflow where a createContainer came in it would error to call it again.

service/gcs/bridge/bridge.go
jterry75: Actually you cant. Golang only has math.Min for floats...... And this is the recommend pattern in Go

service/gcs/bridge/bridge.go
jterry75: Same answer. This is a *multi* call protection. If we aren't worried about that I can remove.

service/gcs/bridge/bridge.go
beweedon: Ah, right.

service/gcs/bridge/bridge.go
```diff
@@ -317,7 +340,55 @@ func (b *Bridge) PublishNotification(n *prot.ContainerNotification) {
...
+	response := &prot.NegotiateProtocolResponse{
+		MessageResponseBase: &prot.MessageResponseBase{
+			ActivityID: request.ActivityID,
+		},
+		Version:      prot.Version{Major: major, Minor: 0, Patch: 0, Metadata: ""},
+		Capabilities: capabilities,
+	}
+
+	// Set our protocol selected version before return.
+	b.protVer = prot.ProtocolVersion(major)
```
beweedon: There's potentially a race condition here if two protocol negotiation/create messages are sent in quick succession. Maybe protect this with a mutex here and in `createContainer`?

jterry75: Spoke offline. Its ok to trust the HCS call chain here I think.

service/gcs/bridge/bridge.go
jterry75: You win. Removing.

service/gcs/bridge/bridge.go
jterry75: You win. Removed

beweedon:
@soccerGB, how's this look?

beweedon:
Hold off on merging this for a bit until we're sure it's ready to be merged.

beweedon:
Merging this now into the multicontainer branch, which will be used to test V4 changes before it's all merged back into master.

service/gcs/bridge/bridge.go
```diff
@@ -365,10 +425,17 @@ func (b *Bridge) createContainer(w ResponseWriter, r *Request) {
...
+	b.protVer = prot.PvV3
 	w.Write(response)
 }
 
 func (b *Bridge) execProcess(w ResponseWriter, r *Request) {
+	if r.Version != prot.PvV3 {
+		w.Error("", gcserr.NewHresultError(gcserr.HrVmcomputeUnsupportedProtocolVersion))
+		return
+	}
+
```
soccerGB: This above checking purposely rejects anything that's not V3.
What protocol version does the RS4 HCS use? V4?

beweedon: That check was removed in a later commit in this PR. Also, yeah, the RS4 HCS will use V4. We're currently working on the V4 stuff in the separate `multicontainer` branch, so that we don't break anything in master until it's all ready to be merged in.

PR #181, opened 2018-01-17
Basic panic logger for gcstools

jhowardmsft:

Writes a paniclog if one of the GCS tools crashes, and turns on client-side (docker) ability to be able to grab the logs back. Will be written as `/tmp/paniclog.<toolname>.pid`

@jterry75 As mentioned


jterry75:
Love it!

PR #182, opened 2018-01-17
Add logging flags for remotefs

jhowardmsft:

@jterry75 Don't ask. I really don't like it, but it works.

### DIFF
--- a/service/gcsutils/gcstools/commoncli/common.go
+++ b/service/gcsutils/gcstools/commoncli/common.go
@@ -5,9 +5,10 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
-	"github.com/sirupsen/logrus"
 	"github.com/docker/docker/pkg/archive"
+	"github.com/sirupsen/logrus"
 
 	"github.com/Microsoft/opengcs/service/gcsutils/fs"
 	"github.com/Microsoft/opengcs/service/gcsutils/libtar2vhd"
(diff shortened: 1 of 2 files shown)


service/gcsutils/gcstools/commoncli/common.go
```diff
@@ -70,19 +71,36 @@ func SetupTar2VHDLibOptions(args ...*string) (*libtar2vhd.Options, error) {
...
+	logrus.SetLevel(level)
+
+	// Add the sub-command to the filename for remotefs. This is really ugly :(
+	// Done to allow mixing both fixed args that remotefs expects for back-compat,
+	// along with flags. Prior to this change, remotefs didn't accept flags at all,
+	// and did no logging. Unfortunately, there are circumstances where two distinct
+	// remotefs commands can execute simultaneously, so we need separate logfiles.
+	filename := *args[0]
+	if os.Args[0] == "remotefs" {
+		if len(flag.Args()[0]) > 1 {
```
jterry75: Oh wow.

jterry75:
I hesitate to say LGTM but LGTM.

gupta-ak:
Hm, couldn't the flags be added to the end of the remotefs cmdline? Like the format would look: `remotefs <CMD> [CMD ARGS] [flags]`. For example, the code sequence for `remotefs stat <path>` just checks if the len(argv) <= 3, so I think it won't reject something like `remotefs stat <path> --logging=debug`.

jhowardmsft:
@gupta-ak No. flags.parse expects its flags to be first unfortunately. When it gets down to the functions, the arguments passed are just those which aren't flags, so the current logic still works.

gupta-ak:
Oh I see. It works differently than `getopt` which lets you put the flags at the end. That is unfortunate :(

jhowardmsft:
So OK to merge?

PR #183, opened 2018-01-18
Translate io.EOF

jhowardmsft:

Without this (and a matching docker fix), ReadFile will fail with the pipe is being closed during its io.Copy as it never gets the EOF when all data has been read.

@jterry75 @gupta-ak


soccerGB:
Do you mind telling us what the "matching docker fix" is?

jhowardmsft:
@soccerGB https://github.com/moby/moby/pull/36051

jterry75:
LGTM

Issue #173, opened 2017-12-01
Fully support GetProperties


beweedon:
Right now, we only return the list of running processes. We should actually support all necessary properties to the extent Windows does. We should also make sure to fill in the OS type as Linux for the properties.

jterry75:
This is now completed.

PR #185, opened 2018-01-20
Return the correct structure from GetProperties

beweedon:
Nothing in the HCS was ever sending a GetProperties message, so this
issue didn't really ever come up. We should fix it now, though.

Addresses #173 for V1 containers, but not V2 containers.


jterry75:
LGTM

PR #186, opened 2018-01-24
Handle valid errors on process CloseStdin()

jhowardmsft:

@darrenstahlmsft Yes, this was the bug. @jterry75 - this fixes the reliability bug building Linuxkit I've been trying to track down.

### DIFF
--- a/client/process.go
+++ b/client/process.go
@@ -84,8 +84,11 @@ func (config *Config) RunProcess(commandLine string, stdin io.Reader, stdout io.
 		}
 
 		// Don't need stdin now we've sent everything. This signals GCS that we are finished sending data.
-		if err := process.Process.CloseStdin(); err != nil {
-			return nil, err
+		if err := process.Process.CloseStdin(); err != nil && !hcsshim.IsNotExist(err) && !hcsshim.IsAlreadyClosed(err) {
+			// This error will occur if the compute system is currently shutting down
+			if perr, ok := err.(*hcsshim.ProcessError); ok && perr.Err != hcsshim.ErrVmcomputeOperationInvalidState {
+				return nil, err
+			}
 		}
 	}
 


client/process.go
```diff
@@ -84,8 +84,15 @@ func (config *Config) RunProcess(commandLine string, stdin io.Reader, stdout io.
 		}
 
 		// Don't need stdin now we've sent everything. This signals GCS that we are finished sending data.
-		if err := process.Process.CloseStdin(); err != nil {
-			return nil, err
+		err := process.Process.CloseStdin()
```
beweedon: **nit:**
```go
if err := process.Process.CloseStdin(); err != nil {
    ...
}
```

client/process.go
```diff
@@ -84,8 +84,15 @@ func (config *Config) RunProcess(commandLine string, stdin io.Reader, stdout io.
...
+		if err != nil {
+			if !hcsshim.IsNotExist(err) && !hcsshim.IsAlreadyClosed(err) {
+				// This error will occur if the compute system is currently shutting down
+				if perr, ok := err.(*hcsshim.ProcessError); ok && perr.Err != hcsshim.ErrVmcomputeOperationInvalidState {
+					return nil, err
+				}
+			}
+			logrus.Debugf("opengcs: RunProcess %s ignoring legitimate error %v on CloseStdin()", commandLine, err)
```
darrenstahlmsft: nit: Do we care about logging this? I think it would just generate extra noise.

jhowardmsft: I only put it in to verify my check was trapping. Will remove.

client/process.go
jhowardmsft: Updating shortly...

client/process.go
```diff
@@ -84,8 +84,15 @@ func (config *Config) RunProcess(commandLine string, stdin io.Reader, stdout io.
 		}
 
 		// Don't need stdin now we've sent everything. This signals GCS that we are finished sending data.
-		if err := process.Process.CloseStdin(); err != nil {
-			return nil, err
+		err := process.Process.CloseStdin()
+		if err != nil {
+			if !hcsshim.IsNotExist(err) && !hcsshim.IsAlreadyClosed(err) {
+				// This error will occur if the compute system is currently shutting down
+				if perr, ok := err.(*hcsshim.ProcessError); ok && perr.Err != hcsshim.ErrVmcomputeOperationInvalidState {
```
darrenstahlmsft: Not for this PR, but we should add this to one of the error checking functions in hcsshim rather than inspecting the errors in consumers.

darrenstahlmsft:
LGTM

jhowardmsft:
@darrenstahlmsft @beweedon Updated.

client/process.go
jhowardmsft: I'm not sure, there's that example in tar2vhd where I think ANY error is legitimate. But, yes, it could be refactored still later.

beweedon:
LGTM

jhowardmsft:
@soccerGB Can you merge? (Not sure why I don't have merge privilege for things under client/ though)

client/process.go
darrenstahlmsft: I mean moving this check into hcsshim.IsAlreadyClosed, as this error is just a symptom of the process already being closed. That way we can ignore it here, and still error in tar2vhd

PR #189, opened 2018-01-30
Properly handle case where GetProperties returns nil on non error

beweedon:


service/gcs/bridge/bridge.go
```diff
@@ -470,10 +470,14 @@ func (b *Bridge) getProperties(w ResponseWriter, r *Request) {
 		return
 	}
 
-	propertyJSON, err := json.Marshal(properties)
-	if err != nil {
-		w.Error(request.ActivityID, errors.Wrapf(err, "failed to marshal properties into JSON: %v", properties))
-		return
+	propertyJSON := []byte("{}")
```
jterry75: What does setting this default help us?

beweedon: This prevents the GCS from returning with the JSON string `"null"` if `GetProperties` is called on a container which has been created, but which hasn't been given its first process yet. This way, the HCS receives back an empty properties response, rather than `"null"` and throwing invalid JSON.

jterry75: Oh the lines of context really killed me on that one. I see later how its used.

jterry75:
LGTM

beweedon:
ping

jterry75:
I already said LGTM but just signed off in case you were waiting on me.

beweedon:
Since this is going into master, we just need @soccerGB to sign off.

Issue #188, opened 2018-01-29
LCOW: App compat: WORKDIR should create the target directory


jhowardmsft:
Moving from https://github.com/moby/moby/issues/36138

@jterry75 As per discussion, should have the same contract as for WCOW where gcs/cexecsvc do the mkdirall (`mkdir -p`) and ignore the error.

Building redmine fails under LCOW unless an explicit RUN mkdir -p <workdir> immediately before the WORKDIR statement.
(Internal tracking VSO#14546486)

PR #190, opened 2018-01-31
Creates the workdir on start

jterry75:
Propertly creates the workdir in the container bundle previous to a start if
that workdir is not '/'

Ref: #188


jterry75:
@jhowardmsft - I will figure out exec tomorrow

PR #192, opened 2018-02-03
Fix int overflow on 32 bit architectures

rbalint:


jterry75:
Code looks good but out of curiosity how did you hit this. This code would never be running on 32-bit architectures to my knowledge.

rbalint:
2018-02-06 1:49 GMT+07:00 Justin <notifications@github.com>:
>
> Code looks good but out of curiosity how did you hit this. This code would never be running on 32-bit architectures to my knowledge.

In Ubuntu we compile packages for all architectures by default and
compilation failed:
https://bugs.launchpad.net/ubuntu/+source/opengcs/+bug/1746697

jterry75:
Ah. LGTM.

PR #194, opened 2018-02-07
Changing bridge to major numbers only

jterry75:


beweedon:
Looks good once you fix the tests!

jterry75:
I am going to update the makefile to actually compile the tests locally. So tired of hitting that lol

service/gcs/prot/protocol.go
```diff
@@ -172,15 +172,6 @@ type GcsCapabilities struct {
 	SendInitialCreateMessage bool
 }
 
-// Version is the struct sent to identify the protocol version of the HCS-GCS
-// communication protocol.
-type Version struct {
```
beweedon: **nit:** Might as well keep the `Version` struct, and just remove the `Patch` and `Metadata` fields, since we'll need `Version` later anyway.

jterry75: I will add it back as `SchemaVersion` because I think its more clear to differentiate them. To may `Version` concepts.

beweedon: Sounds good to me!

PR #205, opened 2018-04-10
Unplug mapped disks from inside the VM

rn:
Currently, when removing mapped disks the disks are unmounted
before the host removes them. The removal from the host is just
a notification and the Linux kernel then performs the clean-up
asynchronously.

If we unplug a larger number of disks this may take some time
and there is no way for the host to know when the clean-up is
finished and new/different disks can be added.

This patch tells the Linux kernel that we are about to unplug
the disks after they were unmounted, allowing it to perform
the clean-up in advance.

This fixes the issue in globalmode where a mkdir fails within
the Linux utility/service VM.

Signed-off-by: Rolf Neugebauer <rolf.neugebauer@docker.com>

### DIFF
--- a/service/gcs/oslayer/realos/realos.go
+++ b/service/gcs/oslayer/realos/realos.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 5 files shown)


rn:
@jhowardmsft @jterry75 I've done about 20 OpenGCS builds in global mode without hitting the `mkdir` or related errors once with this patch. Deleteing/unplugging the device from the within the VM prior to unplugging it from the host side is definitely the correct thing to do.

If you want to try for yourself, here is a `initrd` with the changes
[initrd.img.zip](https://github.com/Microsoft/opengcs/files/1895326/initrd.img.zip)

soccerGB:
Thanks for root causing this issue.
Those changes look good to me.

service/gcs/oslayer/realos/realos.go
```diff
@@ -184,6 +185,18 @@ func (o *realOS) Unmount(target string, flags int) (err error) {
...
 	return nil
 }
+func (o *realOS) UnplugSCSIDisk(scsiID string) (err error) {
+	f, err := os.OpenFile(filepath.Join("/sys/bus/scsi/devices", scsiID, "delete"), os.O_WRONLY, 0644)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	defer f.Close()
+
+	if _, err := f.Write([]byte("1\n")); err != nil {
```
jhowardmsft: Just trying to get my head around what this does - writing `1` here? But LGTM otherwise.

rn: the files in `/sys` are magic :) For some of them if you write anything to them stuff will happen. For the `delete` file in the `scsi` hierarchy writing anything to the file will case the unplugging. It is customary to do something like `echo 1 > /sys/..../delete`, which basically this code does.

PR #207, opened 2018-06-04
Stop calling kernel bootx64.efi. It confuses everybody!

jhowardmsft:

Now that 4 people have got thoroughly confused by the naming of this file, let's just call it what it really is. 

@rn Once PR'd into moby/moby, it will affect deployment through D4W.

### DIFF
--- a/client/config.go
+++ b/client/config.go
@@ -110,7 +110,7 @@ func ParseOptions(options []string) (Options, error) {
 		rOpts.Vhdx = filepath.Join(rOpts.KirdPath, `uvm.vhdx`)
 	}
 	if rOpts.KernelFile == "" {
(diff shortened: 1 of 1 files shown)


jhowardmsft:
@jterry75 PTAL

rn:
@jhowardmsft thanks for the heads up. Could you ping me on the moby/moby PR as well?

jterry75:
This change is fine. I assume upstream you are going to have the code that either renames this on the fly or never passes "" here? Otherwise all the files that are currently on peoples machines will not work.

jhowardmsft:
@jterry75 - yes, I'll take care of upstream to moby/moby and sending out the breaking change to internal users using master.dockerproject.org. I pinged @rn above to sync the D4W change.

PR #214, opened 2018-06-08
Use a single Makefile to build initrd and contents

jstarks:


jstarks:
Yes -- bsdtar can convert archives between types without extracting them, which means you don't need root to build now (as long as you have an existing base.tar.gz).

jstarks:
Please hold merging until I get travis working well with this new build model.

jstarks:
OK, this is ready to go. @jterry75 is going to fix the failing tests.

PR #217, opened 2018-06-13
Fix makefile

jterry75:


jstarks:
I don't think this is what you want... You need a base.tar.gz with a valid root file system and various dependencies (including runc) to end up with a usable initrd.

PR #218, opened 2018-06-13
Return suported versions is GcsCapabilities

jterry75:
1. Now returns the fully supported list in GcsCapabilities


service/gcs/bridge/bridge.go
```diff
@@ -28,6 +28,16 @@ var capabilities = prot.GcsCapabilities{
...
 	SendHostStartMessage:    false,
 	HVSocketConfigOnStartup: false,
+	SupportedSchemaVersions: []prot.SchemaVersion{
+		prot.SchemaVersion{
+			Major: 1,
+			Minor: 0,
+		},
+		prot.SchemaVersion{
+			Major: 2,
+			Minor: 0,
```
jstarks: Should this be 2, 1? That's what it is in Windows RS5.

PR #220, opened 2018-06-14
Build rootfs2vhd and run it

jhowardmsft:

@jterry75 PTAL

Cross-compiles the rootfs2vhd.exe tool in the opengcs LCOW container, and the build script runs it if on an RS5+ machine. Currently any generated error is ignored, as the tool still has a dependency on New-VHD which may not be present (yes, it's on my list of TODOs....)


jterry75:
LGTM

PR #222, opened 2018-06-15
Make logs readable!

jhowardmsft:

@jterry75 PTAL

The current logging is crazy hard to read. There's no point always logging the line of code and file in logrus where the log is made!

```time="2018-06-15T10:07:19-07:00" level=info msg="/go/src/github.com/Microsoft/opengcs/vendor/github.com/sirupsen/logrus/entry.go:97 github.com/Microsoft/opengcs/vendor/github.com/sirupsen/logrus.Entry.log() bridge: response sent: '{\"Result\":0,\"ActivityId\":\"00000000-0000-0000-0000-000000000000\",\"ExitCode\":0}' to HCS" time="2018-06-15T10:07:18Z" vm.time="2018-06-15T10:07:18Z"```

With this change, it's now:

```
time="2018-06-15T11:39:04-07:00" level=info msg="bridge: response sent: '{\"Result\":0,\"ActivityId\":\"00000000-0000-0000-0000-000000000000\",\"ProcessId\":111}' to HCS" time="2018-06-15T11:39:02Z" vm.time="2018-06-15T11:39:02Z"
```

Next to remove is vm.time= as that the same as time=, but that's in the hcsshim repo.

It also adds an attempt to kill the rootfs2vhd container in the build script. Should really be in a separate PR, but while I had it...

### CODE REVIEW

File: service/gcs/main.go [hunk 1]
```diff
@@ -59,10 +58,6 @@ func main() {
 		logrus.Fatal(err)
 	}
 
-	if level == logrus.DebugLevel {
```
jterry75: I would prefer we just stop passing DebugLog level in init by default. That way we can still get debug logs in vsockexec if we pass this verbosity.

PR #223, opened 2018-06-15
Revert "Build rootfs2vhd and run it"

jstarks:
If you want to update `build.ps1` to pull random stuff from hcsshim, fine, but don't pollute the Makefile with this.


jhowardmsft:
OK, will do through build.ps1 and run a container instead.

PR #224, opened 2018-06-15
Add Makefile rule to build rootfs.vhd

jstarks:
This rule requires root or something like a privileged Docker container in order to run.

We can update build.ps1 to use this once Docker LCOW supports configuring capabilities and the device cgroup rules.


jterry75:
LGTM

PR #226, opened 2018-06-20
Adds support for the RuntimeOsType Capability

jterry75:


beweedon:
Looks good to me! And the type used for `OtherCapabilities` can be changed to something other than `interface{}` once you start using it. 😃

PR #235, opened 2018-07-04
V2 changes to modifysettings type

jterry75:


jterry75:
@beweedon - I didnt understand from the notification when you said "Request" needs to accept both. Are you saying we will no longer be using the "v2Request" for v2 style calls?

beweedon:
Yeah. As it stands now, `v2Request` is only used for RS4 v2 GCSes, and since as I understand it you don't intend opengcs to support RS4, you should just be able to avoid using `v2Request` at all. The `Request` field may contain a V1 or a V2 request (or, since in RS5 it's totally pass-thru from the client, you can define your own V2 request if necessary).

jterry75:
@beweedon - Ok this should support `v2Request` as well as a V2 request in `Settings` on the GCS side. Can you give me a quick once over?
@jhowardmsft - We will need to update `hcsshim` to the `GuestRequest` work now that Ben has checked this in

Issue #234, opened 2018-06-27
v1 clients aren't trapping a shutdown failure correctly post multi-container


jhowardmsft:
@jterry75 As mentioned offline. I can't see the obvious change, but docker using v1 client is broken for a simple build using initrd.img from master

eg 

```

PS E:\docker\build\36793> docker build --no-cache --platform=linux -t testcachelinux 1
Sending build context to Docker daemon  2.048kB
Step 1/2 : FROM alpine
 ---> 3fd9065eaf02
Step 2/2 : RUN echo 1 > /test.txt
 ---> Running in 4cdc545dee6f
The command '/bin/sh -c echo 1 > /test.txt' returned a non-zero code: 4294967295: failed to shutdown container: container 4cdc545dee6fbcca7a9b4f9d238c6a64edc0260f9e7f855f4ff5f5ecad8023cc encountered an error during Shutdown: failure in a Windows system call: Unspecified error (0x80004005)
PS E:\docker\build\36793>
```

```
DEBU[2018-06-27T15:45:05.976009700-07:00] HCSShim::Process::properties succeeded processid=128, properties={"ProcessId":128,"Exited":true,"ExitCode":0,"LastWaitResult":0}
DEBU[2018-06-27T15:45:05.977011600-07:00] HCSShim::Process::ExitCode succeeded processid=128 exitCode=0
DEBU[2018-06-27T15:45:05.977011600-07:00] HCSShim::Process::Close processid=128
DEBU[2018-06-27T15:45:05.977011600-07:00] HCSShim::Process::Close succeeded processid=128
DEBU[2018-06-27T15:45:05.978000300-07:00] HCSShim::Container::Shutdown id=4cdc545dee6fbcca7a9b4f9d238c6a64edc0260f9e7f855f4ff5f5ecad8023cc
[...]

PR #237, opened 2018-07-09
Fixes v1 shutdown failure code

jterry75:
1. Fixes an issue where when signaling a container either via Shutdown
or Terminate the HCS previously would return an error that indicated the
Compute System was already stopped. We now have the gcs return this
HRESULT directly to honor the existing hcsshim behavior for determining
if the signal error is a real one or not.


jhowardmsft:
Verified. Fixes https://github.com/Microsoft/opengcs/issues/234

PR #238, opened 2018-07-10
Remove ProcessV2 Version property requirement

jterry75:
Removes the SchemaVersion property requirement from the ProcessV2 schema
and gathers the version number from the ContainerID itself that was used
at create time.


jhowardmsft:
CI failure looks legit...

jterry75:
@jhowardmsft - PTAL. Was a valid unit test failure but also not important. Fixed now

Issue #246, opened 2018-08-13
V2 exec process should consider UVMId


jterry75:
For V2 exec process against the host we really should consider
```go
if params.IsExternal {
  // v1 external process
} else if request.ContainerID == gcspkg.UVMContainerID {
  // V2 external process
  // use oci.Process 
}
... // Container processes
```

PR #249, opened 2018-08-15
Supports V2 exec external process with a V2 schema

jterry75:
Adds support for a V2 exec external process that uses an oci.Process
spec rather than the v1 schema of cmd/args.

Resolves: #246


jterry75:
@jhowardmsft - PTAL

PR #250, opened 2018-08-15
Fixing time.Duration casts

jterry75:


jterry75:
@jiria - This is what I was asking for. I think we can remove #250 and take this instead. Do you see an issue with that?

jterry75:
Oh I see the issue. This client code is not actually built in this repo. Hang on will fix this.

PR #262, opened 2018-10-15
Return GuestDefinedCapabilities with SignalProcess support

jterry75:

### DIFF
--- a/service/gcs/bridge/bridge.go
+++ b/service/gcs/bridge/bridge.go
@@ -41,6 +41,9 @@ var capabilities = prot.GcsCapabilities{
 		},
 	},
 	RuntimeOsType: prot.OsTypeLinux,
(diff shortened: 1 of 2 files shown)


service/gcs/bridge/bridge.go
```diff
@@ -41,6 +41,9 @@ var capabilities = prot.GcsCapabilities{
 		},
 	},
 	RuntimeOsType: prot.OsTypeLinux,
+	GuestDefinedCapabilities: prot.GcsGuestCapabilities{
+		SignalProcessSupported: true,
```
jhowardmsft: Are we going to support NamespaceAddRequestSupported at a later date?

jterry75: Yes we dont support it yet.

jhowardmsft: SGTM

PR #265, opened 2018-12-05
Remove tar2vhd from the OpenGCS

jterry75:
tar2vhd is a tool that converted a Docker layer tar to a vhd by use of
streaming the tar to a LCOW UtilityVM and extracting the tar to an ext4
filesystem and streaming back the vhd. This tool has been replaced by a tool
in github.com/Microsoft/hcsshim/cmd/tar2ext4 that can do all of this locally
on a host without the need of a UtilityVM and is much faster. Use this tool
instead moving forward.


jterry75:
hcsshim FYI: https://github.com/Microsoft/hcsshim/pull/412

jterry75:
@jhowardmsft - Can you test this with docker to make sure there is no regression? CD doesnt exercise any of the changed code. I will do the tests on this initrd though just to make sure

jterry75:
Oh awesome! Ok I will check it in then.

PR #267, opened 2019-01-04
Implement V2 Network HotAdd/Remove

jterry75:
Adds the ability to do a HotAdd/Remove to either a pre-created container or
to an existing container that already holds a network namespace.

### DIFF
--- a/service/gcs/core/gcs/networking.go
+++ b/service/gcs/core/gcs/networking.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/Microsoft/opengcs/service/gcs/oslayer"
 	"github.com/Microsoft/opengcs/service/gcs/prot"
 	"github.com/Microsoft/opengcs/service/gcs/runtime"
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


jterry75:
**DO NOT MERGE** - Looking for early feedback. @madhanrm - Does this work as you suggested to use `NatEnabled: a.IPAddress != ""`. 

Thanks

jterry75:
@jhowardmsft - What the heck is the vendor process here? This is not `vndr` and its not `Godep`. We need to fix this

jhowardmsft:
@jterry75 - Going to hazard a guess that it's using https://github.com/kardianos/govendor. @soccerGB may be able to confirm as he commited the original vendor.json file.

service/gcs/core/gcs/networking.go
```diff
@@ -18,7 +19,7 @@ import (
 // namespace and configures it there.
 func (c *gcsCore) configureAdapterInNamespace(container runtime.Container, adapter prot.NetworkAdapter) error {
 	id := adapter.AdapterInstanceID
-	interfaceName, err := c.instanceIDToName(id)
+	interfaceName, err := instanceIDToName(c.OS, id)
```
jhowardmsft: I'm confused - why is the OS needed here?

jterry75: Because `instanceIDToName` issued a `c.OS.ReadDir` but we need to call this from V2 which doesn't have `(c *gcsCore)` but does have `oslayer.OS`. So instead we just pass the part of `*gcsCore` we need and then V2 can do the same.

jterry75:
@jhowardmsft - PTAL verified this works.

PR #268, opened 2019-01-10
Set the local loopback interface up on init

jterry75:

### DIFF
--- a/init/init
+++ b/init/init
@@ -56,6 +56,8 @@ do
         esac
 done < /proc/cgroups
 
+/sbin/ip link set lo up
+
 if [ $# -eq 0 ]; then
         # Run gcs in the background
         /bin/gcs -loglevel debug -logfile=/tmp/gcs.log &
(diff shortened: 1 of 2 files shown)


init/init
```diff
@@ -56,6 +56,8 @@ do
         esac
 done < /proc/cgroups
 
+/sbin/ip link set lo up
```
jterry75: Does anyone have issues doing it this way? It seems that for POD's we need the loopback interface to be up which is required to be done in the host namespace and rather than coding that in GCS we can just do it here.

jterry75:
Before change:
```sh
/ # ip addr
1: lo: <LOOPBACK> mtu 65536 qdisc noop state DOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
```
After change:
```sh
/ # ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
```

Issue #269, opened 2019-01-15
init does too little and too much


jstarks:
`init` does too little: PID 1 is responsible for reaping zombie processes, but `init` just execs `gcs`, which does not reap zombie processes. `init` should probably fork and exec `gcs` and remain as pid 1, reaping zombie processes all the while.

`init` does too much: it mounts a bunch of file systems to conform with some distro requirements, but I doubt we need much of it. E.g. we probably don't need to mount /dev/mqueue or /dev/shm.

`init` also depends on `sh`, which would not otherwise be necessary (except for debugging).

Probably `init` should be rewritten in C. Its responsibilities are similar to that of Yelp's dumb-init, but it needs to do some early initialization. We could just do the early initialization and exec dumb-init, or we could reimplement the whole thing (which is tiny, especially since we don't need the signal forwarding behavior).

PR #270, opened 2019-01-16
init: Replace with C implementation

jstarks:


jstarks:
Fixes #269.

jstarks:
Fixed issue with LCOWv1 by updating PATH handling. I haven't tested LCOWv2, just uvmboot.

jterry75:
Perf seems fine (no regressions and actually an improvement which is a plus)

PR #271, opened 2019-01-16
Add logrus.WithFields for all V2 calls

jterry75:

### DIFF
--- a/service/gcs/bridge/bridge.go
+++ b/service/gcs/bridge/bridge.go
@@ -43,7 +43,7 @@ var capabilities = prot.GcsCapabilities{
 	RuntimeOsType: prot.OsTypeLinux,
 	GuestDefinedCapabilities: prot.GcsGuestCapabilities{
 		NamespaceAddRequestSupported: true,
-		SignalProcessSupported: true,
+		SignalProcessSupported:       true,
 	},
 }
 
@@ -98,7 +98,10 @@ func (mux *Mux) Handle(id prot.MessageIdentifier, ver prot.ProtocolVersion, hand
 	}
 
 	if _, ok := mux.m[id][ver]; ok {
-		logrus.Infof("bridge: overwriting bridge handler for type: %v, version: %v", id, ver)
+		logrus.WithFields(logrus.Fields{
+			"message-type":     id.String(),
+			"protocol-version": ver,
+		}).Warn("opengcs::bridge - overwriting bridge handler")
 	}
 
 	mux.m[id][ver] = handler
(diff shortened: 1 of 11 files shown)


service/gcs/core/gcs/storage.go
```diff
@@ -185,14 +190,14 @@ func scsiControllerLunToName(osl oslayer.OS, controller, lun uint8) (string, err
 
 // scsiLunToName finds the SCSI device with the given LUN. This assumes
 // only one SCSI controller.
-func scsiLunToName(osl oslayer.OS, lun uint8) (string, error) {
-	return scsiControllerLunToName(osl, 0, lun)
+func (c *gcsCore) scsiLunToName(lun uint8) (string, error) {
```
jterry75: This was V1 only. Marking that explicitly here. If we want this a separate change I can do that.

service/gcs/core/gcs/storage.go
```diff
@@ -185,14 +190,14 @@ func scsiControllerLunToName(osl oslayer.OS, controller, lun uint8) (string, err
...
+	return scsiControllerLunToName(c.OS, 0, lun)
 }
 
 // deviceIDToName converts a device ID (scsi:<lun> or pmem:<device#> to a
 // device name (/dev/sd? or /dev/pmem?).
 // For temporary compatibility, this also accepts just <lun> for SCSI devices.
-func deviceIDToName(osl oslayer.OS, id string) (device string, pmem bool, err error) {
+func (c *gcsCore) deviceIDToName(id string) (device string, pmem bool, err error) {
```
jterry75: This was V1 only. Marking that explicitly here. If we want this a separate change I can do that.

service/gcsutils/fs/ext4.go
```diff
@@ -134,32 +134,25 @@ func (e *Ext4Fs) CleanupSizeContext() error {
...
-	if err != nil {
-		logrus.Infof("running mkfs.ext4 failed with ... (%s)", err)
-	}
-
-	return err
-}
-
-// MakeBasicFileSystem just creates an empty file system on the given file using
-// the default settings.
-func (e *Ext4Fs) MakeBasicFileSystem(file *os.File) error {
```
jterry75: Unused code.

jterry75:
I have verified all this output and logging. Looks good for now.

service/gcs/bridge/bridge.go
```diff
@@ -621,12 +668,19 @@ func (b *Bridge) signalContainer(w ResponseWriter, r *Request, signal oslayer.Si
 		return
 	}
 
+	fields := logrus.Fields{
```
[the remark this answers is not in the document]
jterry75: Fair enough.

service/gcs/core/gcs/uvm.go
```diff
@@ -466,21 +490,27 @@ func (c *Container) AddNetworkAdapter(o oslayer.OS, a *prot.NetworkAdapterV2) er
 		}
 		break
 	}
-	logrus.Debugf("Waited: %v for the network adapter to show up", time.Now().Sub(start))
+	log.Data["adapter-wait-time-ns"] = time.Now().Sub(start).Nanoseconds
```
kevpar: Could use time.Since(start) if you prefer

PR #273, opened 2019-01-31
Implement /etc/resolv.conf bind mount for sandbox/workload containers

jterry75:
Creates an /etc/resolv.conf file per sandbox/standalone container with a
network namespace and bind mounts that into the running container. For a
Kubernetes sandbox container will bind mount the same /etc/resolv.conf file
into all Workload containers with the same shared namespace.

### DIFF
--- a/service/gcs/core/gcs/networking.go
+++ b/service/gcs/core/gcs/networking.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/Microsoft/opengcs/service/gcs/prot"
 	"github.com/Microsoft/opengcs/service/gcs/runtime"
@@ -19,15 +20,14 @@ import (
 // configureAdapterInNamespace moves a given adapter into a network
 // namespace and configures it there.
 func (c *gcsCore) configureAdapterInNamespace(container runtime.Container, adapter prot.NetworkAdapter) error {
-	id := adapter.AdapterInstanceID
-	interfaceName, err := instanceIDToName(id)
+	interfaceName, err := instanceIDToName(adapter.AdapterInstanceID, false)
 	if err != nil {
 		return err
 	}
 	nspid := container.Pid()
 	cfg, err := json.Marshal(adapter)
 	if err != nil {
-		return errors.Wrapf(err, "failed to marshal adapter struct to JSON for adapter %s", id)
+		return errors.Wrapf(err, "failed to marshal adapter struct to JSON for adapter %s", adapter.AdapterInstanceID)
 	}
 
 	out, err := exec.Command("netnscfg",
@@ -71,7 +71,7 @@ func (c *gcsCore) configureAdapterInNamespace(container runtime.Container, adapt
 // method in the future.
 func generateResolvConfFile(resolvPath string, adapter prot.NetworkAdapter) error {
 	logrus.WithFields(logrus.Fields{
-		"adapterID":            adapter.AdapterInstanceID,
+		"adapterInstanceID":    adapter.AdapterInstanceID,
 		"resolvPath":           resolvPath,
 		"host-dns-server-list": adapter.HostDNSServerList,
 		"host-dns-suffix":      adapter.HostDNSSuffix,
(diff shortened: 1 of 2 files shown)


service/gcs/core/gcs/uvm.go
```diff
@@ -96,6 +100,80 @@ func (h *Host) CreateContainer(id string, settings *prot.VMHostedContainerSettin
...
+				}).Info("opengcs::CreateContainer - Generate sandbox resolv.conf")
+
+				// TODO: Can we ever have more than one nic here at start?
+				td, err := ioutil.TempDir("", "")
+				if err != nil {
+					return nil, errors.Wrap(err, "failed to create tmp directory")
+				}
+				resolvPath := filepath.Join(td, "resolv.conf")
+				adp := netopts.Adapters[0]
+				adapterID, err := instanceIDToName(h.osl, adp.ID, true)
```
kevpar: I think ```adapterName``` would be a better name here.

kevpar: Or maybe not, is the terminology that ```eth0``` is the name, and the GUID is the ID? It just seems weird we're calling a function called ```instanceIDToName``` and assigning its output to something named ```adapterID```.

service/gcs/core/gcs/uvm.go
```diff
@@ -96,6 +100,80 @@ func (h *Host) CreateContainer(id string, settings *prot.VMHostedContainerSettin
...
+	var networkNamespace string
+	if settings.OCISpecification.Windows != nil &&
+		settings.OCISpecification.Windows.Network != nil &&
+		settings.OCISpecification.Windows.Network.NetworkNamespace != "" {
+		networkNamespace = strings.ToLower(settings.OCISpecification.Windows.Network.NetworkNamespace)
+	}
+	if networkNamespace != "" {
+		if isSandboxOrStandalone {
+			// We have a network namespace. Generate the resolv.conf and add the bind mount.
+			netopts := h.netNamespaces[networkNamespace]
```
kevpar: Should we check if the map actually has a value for this key before using it? Otherwise we'll panic when accessing netopts.Adapters below if the map doesn't have it.

service/gcs/core/gcs/networking.go
```diff
@@ -109,20 +109,43 @@ func generateResolvConfFile(o oslayer.OS, resolvPath string, adapter prot.Networ
...
+		deviceDirs []os.FileInfo
+		err        error
+	)
+	start := time.Now()
+	for {
+		deviceDirs, err = o.ReadDir(filepath.Join("/sys", "bus", "vmbus", "devices", id, "net"))
+		if err != nil {
+			if wait {
+				if os.IsNotExist(errors.Cause(err)) {
+					<-time.After(10 * time.Millisecond)
```
jstarks: For sync sleeping, using time.Sleep

jterry75: Good call

jterry75:
Ping. I need to check this in can we review?

service/gcs/core/gcs/uvm.go
```diff
@@ -96,6 +100,76 @@ func (h *Host) CreateContainer(id string, settings *prot.VMHostedContainerSettin
...
+				Destination: "/etc/resolv.conf",
+				Type:        "bind",
+				Source:      netopts.ResolvPath,
+				Options:     []string{"bind", "ro"},
+			})
+		}
+	}
+
+	// Clear the windows section of the config
+	settings.OCISpecification.Windows = nil
```
jhowardmsft: Is this just to indicate that you've done the conversion? And not strictly necessary?

jhowardmsft:
Seems sane. Have you verified it doesn't break docker and LCOW v1 networking?

service/gcs/core/gcs/uvm.go
jterry75: Yea. I mostly just want to remove the section as we really should not be passing it to `runc create` as its not a valid Linux config. But they dont seem to crash if we do. Just seems safer to remove.

jterry75:
> Seems sane. Have you verified it doesn't break docker and LCOW v1 networking?

No. But it doesn't touch V1 networking so it shouldn't. I will do a sanity check.

jterry75:
There are no changes. Just rebased because of other conflicting merges. Checking in so we can get an updated official build.

PR #275, opened 2019-02-11
Remove MockOS/RealOS and Oslayer

jterry75:


jterry75:
Need to build/test and make sure the tests don't now fail for odd reasons. Don't merge yet.

service/gcs/bridge/bridge.go
```diff
@@ -652,16 +651,16 @@ func (b *Bridge) execProcess(w ResponseWriter, r *Request) {
 }
 
 func (b *Bridge) killContainer(w ResponseWriter, r *Request) {
-	b.signalContainer(w, r, oslayer.SIGKILL)
+	b.signalContainer(w, r, syscall.SIGKILL)
```
jstarks: It seems that `unix` is the preferred package instead of `syscall` these days.

service/gcs/runtime/runc/runc.go
```diff
@@ -383,39 +383,41 @@ func (r *runcRuntime) pidMapToProcessStates(pidMap map[int]*runtime.ContainerPro
...
+		return -1, errors.Wrapf(err, "failed to find process %d", pid)
 	}
 	state, err := process.Wait()
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed waiting on process %d", pid)
+		return -1, errors.Wrapf(err, "failed waiting on process %d", pid)
 	}
-	return realos.NewProcessExitState(state), nil
+
+	return state.Sys().(syscall.WaitStatus).ExitStatus(), nil
```
jstarks: In recent Go versions, they added a function to get the exit code directly without this cast. It might be Go 1.12, though, so maybe we can't use it yet.

jterry75: Yea its in 1.12

service/gcs/bridge/bridge.go
jterry75: Agreed. And this isnt multiplat so I will switch.

service/gcs/core/gcs/uvm.go
```diff
@@ -555,17 +553,15 @@ func newProcess(c *Container, spec *oci.Process, process runtime.Process, pid ui
...
 			logrus.WithFields(logrus.Fields{
 				"cid":           c.id,
 				"pid":           pid,
 				logrus.ErrorKey: err,
 			}).Error("opengcs::Process - failed to wait for runc process")
-			p.exitCode = -1
-		} else {
-			p.exitCode = state.ExitCode()
 		}
+		p.exitCode = exitCode
```
jhowardmsft: Won't that now give a zero exit code if Wait() fails now?

jhowardmsft: Oh, you changed Wait(). Not sure I like that much as it's kind of prone to error down the road, but OK.

jterry75: I can return `0, err` for error cases. Do you want me to do that? It seemed like 0 was a bad default value because if `err` is ignored `0` is success exit. But I am totally fine putting it back.

jhowardmsft: Nah, it's fine. Both are equally valid.

PR #276, opened 2019-02-11
devicemapper: Code to create and destroy dm devices

jstarks:


jstarks:
Not sure if the package path is right. I don't understand the current package layout in this repo.

jterry75:
So if I understand correctly we would create the mapper with the initial targets that are passed in. This would be 1 per host on `/dev/pmem1`. How do I `HotAdd/Remove` a mapping?

jstarks:
No, targets is a list of ranges for a single block device. Currently, you will always want to pass a single linear target for each device, and you'll create one device per layer.

So you hot add/remove with CreateDevice and RemoveDevice.

jterry75:
Ah makes 100% more sense

PR #277, opened 2019-02-14
Update vendor

jterry75:


jterry75:
@jhowardmsft - Can you make a private build of this and verify LCOW v1 in Docker? I will do this for V2

PR #281, opened 2019-03-04
Unix line endings on .c files

jhowardmsft:

@jterry75 Can you approve/merge? It breaks vendoring in docker as init.c is being pulled in now :(

Just a dos2unix on the two files, no other changes.


jterry75:
We need this on all go files too. Looks like the CI wasn't checking..... UGH is right.

PR #286, opened 2019-03-11
Change HotRemove NIC to warning instead of error

jterry75:
We don't actually fully support HotRemove NIC from namespace but since
this only happens on teardown we likely don't need it. For now just log
this as a warning rather than returning an error because we do
actually try to safe remove it from the host side on shutdown which is
returning an error that is silently ignored.


kevpar:
I don't fully understand the implications of this. Are there cases we need to be careful about where the caller could take some action assuming the interface was actually removed, where this could cause a problem?

jterry75:
Today we only call remove when the POD is being torn down. So we see these failures on the removal path. Technically CNI is supposed to support NIC HotAdd, HotRemove from namespace but we dont today so this is just to avoid sending down the error. If we support HotAdd/HotRemove on Windows we need to fix the GCS here.

PR #294, opened 2019-04-17
Fix errors returned by GCS to align with expectations in the HCS

jterry75:
- Remove unused error types
- Return HR ERROR_NOT_FOUND after process exit
- Return ERROR_VMCOMPUTE_SYSTEM_NOT_FOUND on runc kill error  …
  - When a container is not found the HCS expects a GCS to return an HRESULT
error ERROR_VMCOMPUTE_SYSTEM_NOT_FOUND to indicate this condition.
- Return ERROR_VMCOMPUTE_SYSTEM_ALREADY_EXISTS on CreateContainer with existing ID


internal/runtime/hcsv2/process.go
```diff
@@ -94,6 +94,9 @@ func newProcess(c *Container, spec *oci.Process, process runtime.Process, pid ui
 }
 
 // Kill sends 'signal' to the process.
+//
+// If the processes has already exited returns `gcserr.HrErrNotFound` by
```
jhowardmsft: nit s/processes/process

jhowardmsft:
Minor nit but LGTM

jterry75:
@jhowardmsft - Made the change. Submitting.

PR #297, opened 2019-04-21
gcs: Prevent GCS crashing when external process exits

jstarks:
Currently, when the GCS launches an external process with a pty, that
pty becomes the controlling terminal for the GCS. This causes the kernel
to deliver a SIGHUP signal when the pty is torn down, which causes the
GCS to exit.

To resolve this, the GCS explicitly opens the terminal with O_NOCTTY so
to avoid making it the controlling terminal. It also now marks the child
as the leader of a new session and has the child make the new pty the
controlling terminal of the child process.

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -282,7 +282,9 @@ int reap_until(pid_t until_pid) {
                 }
                 return WEXITSTATUS(status);
             }
-            fputs("child exited by signal\n", stderr);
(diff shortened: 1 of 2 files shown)


jstarks:
This only hits when 1. you create an external process with a tty and 2. GCS is launched without a controlling tty. The latter only happens when GCS is launch directly from init instead of from sh, which happens when you enable the console pipe.

PR #299, opened 2019-05-01
Makefile: append to archive manually

jstarks:
The bsdtar append step do not interoperate well with some archives or
with our tar2ext4 tool. Extract and rearchive files rather than trying
to append or convert in place.

This strips any ownership information from the initrd image, so all
files will be owned by root. This should be acceptable because /dev is
always mounted as devtmpfs, and no other files should need special
owners.


jterry75:
@jstarks - Ok this one is a real issue. It cant find hack/appendcpio.sh

jstarks:
OK fixed.

PR #300, opened 2019-05-01
vsockexec: Remove dependency on Linux headers

jstarks:
This makes it possible to build easily with musl-gcc.


vsockexec/vsockexec.c
```diff
@@ -16,6 +15,19 @@ static const int tcpmode = 1;
 static const int tcpmode;
 #endif
 
+struct sockaddr_vm {
```
jterry75: What are headers even good for right...

PR #301, opened 2019-05-01
Return proper ContainerExit result

jterry75:
1. Fixes a bug where we were returning an exit code in the ContainerExit.Result
field which was incorrect. This is an HRESULT value of the Wait Result not the
exit code which is returned at Process.Wait.

2. Implements v2 to handle Graceful/Forced exit instead of just Unexpected when
the client is responsible for the SIGTERM/SIGKILL that stops the container.

### DIFF
--- a/internal/runtime/hcsv2/container.go
+++ b/internal/runtime/hcsv2/container.go
@@ -29,6 +29,9 @@ type Container struct {
 	container   runtime.Container
 	initProcess *Process
 
+	etL      sync.Mutex
+	exitType prot.NotificationType
+
 	processesMutex sync.Mutex
 	processesWg    sync.WaitGroup
 	processes      map[uint32]*Process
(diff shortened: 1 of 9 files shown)


internal/runtime/hcsv2/process.go
```diff
@@ -7,6 +7,8 @@ import (
 	"sync"
 	"syscall"
 
+	"github.com/Microsoft/opengcs/service/gcs/prot"
+
```
jstarks: nit: extra space

jterry75:
27th try's the charm! I need to get to where we can run the tests locally lol. 

@jstarks - You good with this?

jstarks:
Is there a race where exitType is modified concurrently with reading it? Need a lock or an atomic?

jterry75:
@jstarks - Updated

internal/runtime/hcsv2/container.go
```diff
@@ -203,3 +213,16 @@ func (c *Container) RemoveNetworkAdapter(id string) error {
 	}).Warning("opengcs::Container::RemoveNetworkAdapter - Not implemented")
 	return nil
 }
+
+// setExitType sets `c.exitType` to the appropriate value based on `signal` if
+// `signal` will take down the container.
+func (c *Container) setExitType(signal syscall.Signal) {
+	c.etL.Lock()
+	c.etL.Unlock()
```
jstarks: missing a defer?

jterry75: Good catch.

PR #302, opened 2019-05-01
Fix build remove stale test

jterry75:


jterry75:
path.Unmount is now smart enough to not fail the operation if it isnt mounted and just does the directory delete. This test doesnt do anything we need.

PR #310, opened 2019-05-13
Makefile: Fixes to run in Debian stretch

jstarks:
This fixes a few limitations that prevent the Makefile from being used
in a Debian stretch environment. It also fixes uses of non-standard Go
paths.

### DIFF
--- a/hack/gomakedeps.sh
+++ b/hack/gomakedeps.sh
@@ -4,6 +4,7 @@
 
 set -e
 
+GO="${GO:-go}"
 target_bin="$1"
 target_pkg="$2"
 shift 2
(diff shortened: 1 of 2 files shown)


hack/gomakedeps.sh
```diff
@@ -4,16 +4,17 @@
...
 
+GO="${GO:-go}"
 target_bin="$1"
 target_pkg="$2"
 shift 2
 fmt=$(cat <<EOF
 {{ if not .Standard }}
 $target_bin: {{\$dir := .Dir}}{{ range .GoFiles }}{{\$dir}}/{{.}} {{end}}
-{{if .Module}}$target_bin: {{.Module.GoMod}}{{end}}
+{{if .Module}}"$target_bin: {{.Module.GoMod}}{{end}}
```
jterry75: I don't understand the single `"`. Did you miss one or was this intentional?

jstarks: Weird. Not sure how that snuck in.

jstarks:
Ugh now it doesn't work with alpine.

PR #311, opened 2019-05-13
Ignore ResizeConsole of init process after exit

jterry75:

### DIFF
--- a/service/gcs/stdio/stdio.go
+++ b/service/gcs/stdio/stdio.go
@@ -316,7 +316,7 @@ func (r *TtyRelay) ResizeConsole(height, width uint16) error {
 	defer r.m.Unlock()
 
 	if r.closed {
-		return errors.New("error resizing console pty is closed")
+		return nil
 	}
 	return ResizeConsole(r.pty, height, width)
 }


service/gcs/stdio/stdio.go
```diff
@@ -291,12 +291,19 @@ func (pr *PipeRelay) closePipes() {
 }
 
 // NewTtyRelay returns a new TTY relay for a given master PTY file.
-func NewTtyRelay(s *ConnectionSet, pty *os.File) *TtyRelay {
-	return &TtyRelay{s: s, pty: pty}
+func NewTtyRelay(s *ConnectionSet, pty *os.File, isInit bool) *TtyRelay {
```
jstarks: I think you're pushing too much business logic down here. The TTY relay shouldn't have to care about the initial process vs subprocesses.

Should this be named something like `ignoreClosedError` or something?

jstarks: Alternatively, should we just always ignore errors?

jterry75: The containerd console has an explicit call to close on the fd when they issue the delete. Our relay automatically closes the fd after a call to wait. So we either need to handle init differently or just ignore the errors in general. I am fine with just ignoring the error on closed in general if you want. But its "most correct" to only ignore it for the init process. But since GCS has different lifetimes it doesnt really map

jstarks: I'm open to either, but this relay's interface should not be tied to the concept of initial vs non initial processes, right? If we wanted to reuse this relay elsewhere, `isInit` would be meaningless.

jterry75: We will just ignore them for all closed tty's then

jstarks: OK that's fine, but I'm not sure if I'm being clear--you can also just rename `isInit` to something that means what it does instead of what scenario it's used in.

PR #312, opened 2019-05-15
gcs: Fill out ErrorMessage string for HCS external bridge

jstarks:
To make it unambiguous which string represents the user-actionable
error, return it in a new field in the bridge response.


jterry75:
Discussed on: https://github.com/microsoft/hcsshim/pull/609

This is good

PR #313, opened 2019-05-15
UnknownMessage for v1 requires E_NOTIMPL

jterry75:


jterry75:
The first of many PR's in unwinding the clock for v2. The others will be done together but this one makes more sense to do on its own.

PR #314, opened 2019-05-15
Separate v1 and v2 bridge interfaces

jterry75:
This change breaks apart the v1 and v2 schema interfaces as well as the v3 and
v4+ bridge protocols. It makes the v4+ protocol opt in so that downlevel
clients running v1 Xenons will stay on v3. This allows us to iterate and change
the model for v4+ without risking breaking downlevel.

### DIFF
--- a/service/gcs/bridge/bridge_v2.go
+++ b/service/gcs/bridge/bridge_v2.go
@@ -1,0 +1,519 @@
+package bridge
+
+import (
+	"encoding/json"
+	"syscall"
+	"time"
+
+	"github.com/Microsoft/opengcs/internal/runtime/hcsv2"
+	"github.com/Microsoft/opengcs/service/gcs/gcserr"
+	"github.com/Microsoft/opengcs/service/gcs/prot"
+	"github.com/Microsoft/opengcs/service/gcs/stdio"
+	"github.com/Microsoft/opengcs/service/libs/commonutils"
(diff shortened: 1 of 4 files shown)


jterry75:
@jstarks / @jhowardmsft - Did my best to not make any improvements/functional changes in this PR and to just do the split operation. Have verified v2 and v1 work as expected.

service/gcs/bridge/bridge_v2.go
```diff
@@ -0,0 +1,520 @@
...
+
+	log := logrus.WithFields(logrus.Fields{
+		"activityID": request.ActivityID,
+		"cid":        request.ContainerID,
+		"signal":     signal,
+	})
+	log.Info("opengcs::bridge::signalContainerV2")
+
+	// V2 added support for sending a signal to the host UVM itself. See if this is targeting
+	// the UVM and then see if its a V2 container ID before falling back to the V1 path.
```
jstarks: nit: update comment

jstarks:
Looks good, probably we should take a cleanup pass after this to remove some awkward branching and other things in the v2 code.

jterry75:
Yep. I did that all at once and then decided its hard to review. The next pr does that

service/gcs/bridge/bridge_v2.go
jterry75: Done.

PR #315, opened 2019-06-05
Fix regression in external process wait on V2

jterry75:


jterry75:
Link: https://github.com/microsoft/hcsshim/pull/623

PR #317, opened 2019-06-06
Simplify the bridge request/return pattern

jterry75:
This change makes it easier for the rquest/response pattern to return a result
or error. This is perferred over the w.Write and w.Error pattern where a
response must be written and then an early return. This makes the code easier
to read and control flow easier to understand.

### DIFF
--- a/service/gcs/bridge/bridge.go
+++ b/service/gcs/bridge/bridge.go
@@ -27,8 +27,8 @@ import (
 
 // UnknownMessage represents the default handler logic for an unmatched request
 // type sent from the bridge.
-func UnknownMessage(w ResponseWriter, r *Request) {
-	w.Error("", gcserr.WrapHresult(errors.Errorf("bridge: function not supported, header type: %v", r.Header.Type), gcserr.HrNotImpl))
+func UnknownMessage(r *Request) (RequestResponse, error) {
+	return nil, gcserr.WrapHresult(errors.Errorf("bridge: function not supported, header type: %v", r.Header.Type), gcserr.HrNotImpl)
 }
 
 // UnknownMessageHandler creates a default HandlerFunc out of the
@@ -39,15 +39,15 @@ func UnknownMessageHandler() Handler {
 
 // Handler responds to a bridge request.
 type Handler interface {
-	ServeMsg(ResponseWriter, *Request)
+	ServeMsg(*Request) (RequestResponse, error)
 }
 
 // HandlerFunc is an adapter to use functions as handlers.
-type HandlerFunc func(ResponseWriter, *Request)
+type HandlerFunc func(*Request) (RequestResponse, error)
 
 // ServeMsg calls f(w, r).
-func (f HandlerFunc) ServeMsg(w ResponseWriter, r *Request) {
-	f(w, r)
+func (f HandlerFunc) ServeMsg(r *Request) (RequestResponse, error) {
+	return f(r)
 }
 
 // Mux is a protocol multiplexer for request response pairs
@@ -86,7 +86,7 @@ func (mux *Mux) Handle(id prot.MessageIdentifier, ver prot.ProtocolVersion, hand
 }
 
 // HandleFunc registers the handler function for the given message id and protocol version.
-func (mux *Mux) HandleFunc(id prot.MessageIdentifier, ver prot.ProtocolVersion, handler func(ResponseWriter, *Request)) {
+func (mux *Mux) HandleFunc(id prot.MessageIdentifier, ver prot.ProtocolVersion, handler func(*Request) (RequestResponse, error)) {
 	if handler == nil {
 		panic("bridge: nil handler func")
 	}
@@ -119,28 +119,31 @@ func (mux *Mux) Handler(r *Request) Handler {
 
 // ServeMsg dispatches the request to the handler whose
 // type matches the request type.
-func (mux *Mux) ServeMsg(w ResponseWriter, r *Request) {
+func (mux *Mux) ServeMsg(r *Request) (RequestResponse, error) {
 	h := mux.Handler(r)
-	h.ServeMsg(w, r)
+	return h.ServeMsg(r)
 }
 
 // Request is the bridge request that has been sent.
 type Request struct {
-	Header  *prot.MessageHeader
+	// Header is the wire format message header that preceeded the message for
+	// this request.
+	Header *prot.MessageHeader
+	// ContainerID is the id of the container that this message cooresponds to.
+	ContainerID string
+	// ActivityID is the id of the specific activity for this request.
+	ActivityID string
+	// Message is the portion of the request that follows the `Header`. This is
+	// a json encoded string that MUST contain `prot.MessageBase`.
 	Message []byte
+	// Version is the version of the protocol that `Header` and `Message` were
+	// sent in.
 	Version prot.ProtocolVersion
 }
 
-// ResponseWriter is the dispatcher used to construct the Bridge response.
-type ResponseWriter interface {
-	// Header is the request header that was requested.
-	Header() *prot.MessageHeader
-	// Write a successful response message.
-	Write(interface{})
-	// Error writes the provided error as a response to the message correlated
-	// with the activity ID passed. If the activity ID is the empty string it
-	// will be translated to an empty guid.
-	Error(string, error)
+// RequestResponse is the base response for any bridge message request.
+type RequestResponse interface {
+	Base() *prot.MessageResponseBase
 }
 
 type bridgeResponse struct {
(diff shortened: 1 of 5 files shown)


jstarks:
This looks good, but instead of returning interface{}, I would suggest returning a type of ```
interface {
    GetBase() *MessageResponseBase
}
```

And then implement GetBase() on *MessageResponseBase to return itself.

This approach ensures that the functions return a response type (could still be the wrong one, of course), and it allows the caller of all these functions to copy the ActivityID field from the request rather than rely on each of the individual code paths to remember to do it.

File: service/gcs/bridge/bridge.go [hunk 1]
```diff
@@ -310,20 +262,22 @@ func (b *Bridge) ListenAndServe(bridgeIn io.ReadCloser, bridgeOut io.WriteCloser
...
-				b.Handler.ServeMsg(wr, r)
-				if !wr.respWritten {
-					logrus.WithFields(logrus.Fields{
-						"message-type": r.Header.Type.String(),
-						"message-id":   r.Header.ID,
-					}).Error("opengcs::bridge - request didn't write response")
+				resp, err := b.Handler.ServeMsg(r)
+				if err != nil {
+					r := &prot.MessageResponseBase{
+						ActivityID: "00000000-0000-0000-0000-000000000000",
```
jstarks: Should be copied from the request?

jterry75: I wish I could. Its part of the json message not the header so if we fail json deserialization we dont have an ActivityID to respond. I have another change where I do the base unmarshal before the function invocation so we can get this as long as we get the message deserialized

jterry75:
> This looks good, but instead of returning interface{}, I would suggest returning a type of ```
> interface {
> GetBase() *MessageResponseBase
> }
> 
> ```
> 
> And then implement GetBase() on *MessageResponseBase to return itself.
> 
> This approach ensures that the functions return a response type (could still be the wrong one, of course), and it allows the caller of all these functions to copy the ActivityID field from the request rather than rely on each of the individual code paths to remember to do it.
> ```

I'm sure I am missing something but how does this help? Not all functions return a `*MessageResponseBase` so having an interface with a `GetBase()` method doesn't help me to eventually serialize the entire structure. It would only allow me to serialize the `*MessageResponseBase` part of the structure.

jstarks:
Take a look at https://github.com/microsoft/hcsshim/blob/master/internal/gcs/protocol.go.

service/gcs/bridge/bridge.go
jterry75: I did this anyways by doing an early deserialization.

jterry75:
> Take a look at https://github.com/microsoft/hcsshim/blob/master/internal/gcs/protocol.go.

Got it thanks! PR Updated

service/gcs/prot/protocol.go
```diff
@@ -646,6 +646,14 @@ type MessageResponseBase struct {
 	ErrorRecords []ErrorRecord `json:",omitempty"`
 }
 
+// GetBase assigns `mrp` if `nil` and returns `mrp`.
+func (mrp *MessageResponseBase) GetBase() *MessageResponseBase {
+	if mrp == nil {
+		mrp = &MessageResponseBase{}
```
jstarks: This gets thrown away--why do it?

service/gcs/prot/protocol.go
```diff
@@ -655,6 +663,12 @@ type NegotiateProtocolResponse struct {
 	Capabilities GcsCapabilities
 }
 
+// GetBase assigns `npr.MessageResponseBase` if `nil` and returns
+// `npr.MessageResponseBase`.
+func (npr *NegotiateProtocolResponse) GetBase() *MessageResponseBase {
```
jstarks: Shouldn't have to reimplement this for each type--Go will automatically expose the methods on any embedded values. However, you'll need to change the embedded type from a pointer to a value (which makes more sense anyway--the response base is not an optional part of the response, so it doesn't need to be a pointer).

service/gcs/bridge/bridge.go
```diff
@@ -301,7 +268,19 @@ func (b *Bridge) ListenAndServe(bridgeIn io.ReadCloser, bridgeOut io.WriteCloser
 					"message-id":   header.ID,
 					"message":      string(message),
 				}).Debug("opengcs::bridge - read message")
-				requestChan <- &Request{header, message, b.protVer}
+				base := prot.MessageBase{}
+				if err := json.Unmarshal(message, &base); err != nil {
+					// Set the ActivityID to empty guid since the json message
+					// is invalid
+					base.ActivityID = "00000000-0000-0000-0000-000000000000"
```
jstarks: If you mark ActivityID as omitempty, then you don't have to set it at all. And indeed, that would fix an existing bug, where if the host omits ActivityID, the guest should not send it back as an empty string in the JSON, it should omit it as well. (I hit this bug while working on the external bridge and had to remove omitempty from ActivityID on the host side. But it should be safe to add here on the guest side.)

service/gcs/bridge/bridge.go
```diff
@@ -310,20 +289,23 @@ func (b *Bridge) ListenAndServe(bridgeIn io.ReadCloser, bridgeOut io.WriteCloser
...
-						"message-id":   r.Header.ID,
-					}).Error("opengcs::bridge - request didn't write response")
+				resp, err := b.Handler.ServeMsg(r)
+				if resp == nil {
+					resp = &prot.MessageResponseBase{
+						ActivityID: r.ActivityID,
+					}
 				}
+				resp.GetBase().ActivityID = r.ActivityID
+				setErrorForResponseBase(resp.GetBase(), err)
```
jstarks: Shouldn't we only do this if err != nil?

service/gcs/bridge/bridge.go
```diff
@@ -310,20 +289,23 @@ func (b *Bridge) ListenAndServe(bridgeIn io.ReadCloser, bridgeOut io.WriteCloser
...
-				b.Handler.ServeMsg(wr, r)
-				if !wr.respWritten {
-					logrus.WithFields(logrus.Fields{
-						"message-type": r.Header.Type.String(),
-						"message-id":   r.Header.ID,
-					}).Error("opengcs::bridge - request didn't write response")
+				resp, err := b.Handler.ServeMsg(r)
+				if resp == nil {
+					resp = &prot.MessageResponseBase{
+						ActivityID: r.ActivityID,
```
jstarks: This assignment is redundant with the one below.

service/gcs/bridge/bridge.go
```diff
@@ -512,34 +489,31 @@ func (b *Bridge) execProcess(w ResponseWriter, r *Request) {
...
 	if err != nil {
-		w.Error(request.ActivityID, err)
-		return
+		return nil, err
 	}
 
-	response := &prot.ContainerExecuteProcessResponse{
+	return &prot.ContainerExecuteProcessResponse{
 		MessageResponseBase: &prot.MessageResponseBase{
 			ActivityID: request.ActivityID,
```
jstarks: unecessary

jstarks: In general we should remove all references to ActivityID except at the dispatch point.

service/gcs/bridge/bridge.go
```diff
@@ -422,12 +403,10 @@ func (b *Bridge) createContainer(w ResponseWriter, r *Request) {
 	// CreateContainerInfo struct.
 	var settings prot.VMHostedContainerSettings
 	if err := commonutils.UnmarshalJSONWithHresult([]byte(request.ContainerConfig), &settings); err != nil {
```
jstarks: As a followup, it would be good to refine this request type to avoid needing multiple unmarshals. If you look again at the protocol.go in HCS, you can see how I defined types to work around the dumb JSON-in-string nesting.

I wonder if you can just take protocol.go.

service/gcs/bridge/bridge.go
jterry75: I reverted the check in the function and did it here.

service/gcs/prot/protocol.go
jterry75: Just switched from pointers as the base message to avoid this.

service/gcs/bridge/bridge.go
jterry75: I thought that was a protocol failure? Are you sure you can have an empty ActivityID?

jstarks: As long as it's omitempty, it will get defaulted correctly on the host (as the empty GUID). If it's not omitempty, then you're in trouble because the host cannot unmarshal empty string as empty GUID.

service/gcs/bridge/bridge.go
jterry75: Yep agreed.

jterry75:
34ths times the charm! @jstarks - PTAL

PR #318, opened 2019-06-10
Forward context logging for all requests/responses

jterry75:
Creates a context chain on all requests and forwards it to all calls for
the duration of the context. This unifies the
message/activity/containerid for all calls and makes tracing easier to
understand.


internal/log/log.go
```diff
@@ -0,0 +1,62 @@
...
+
+type Log struct {
+	op    string
+	start time.Time
+
+	level logrus.Level
+	L     *logrus.Entry
+}
+
+func (l *Log) Done() {
```
jterry75: @kevpar - Kinda took both approaches here. Have both the "start" message in `NewContext` and I log the `duration` in the end message. Really all we need to do is see the end message in automated logging. It will have the "error" if present as well as the timestamps of the operation

jterry75:
@kevpar - PTAL

kevpar:
At this point it seems like we might want to just adopt the OC API instead of building our own span support. I guess there's no real need for OC here, but if we start using it in other places it makes sense to try to adopt it everywhere. Do you think that's worth considering, or that we would want to potentially replace this with OC later on, or keep it entirely separate?

jterry75:
I'm happy to just use OC, I just didn't know how that would solve our `logrus.` problem. Maybe I should make the `log` package more like `log.G(ctx)` from containerd and pull the span from `ctx` and forward the correlation info that way?

kevpar:
> I'm happy to just use OC, I just didn't know how that would solve our `logrus.` problem. Maybe I should make the `log` package more like `log.G(ctx)` from containerd and pull the span from `ctx` and forward the correlation info that way?

Yeah, I think that's probably better. It aligns more closely with how we want to tackle tracing in containerd. My PR on containerd is [here](https://github.com/containerd/containerd/pull/3303) if that's a useful reference.

jterry75:
@kevpar - PTAL. Fixed the break in the unit tests. Also moved the unused v2 funct to the v1 only namespace to make it easier for a clean break

internal/oc/exporter.go
```diff
@@ -0,0 +1,61 @@
...
+	baseEntry.Data["parentSpanID"] = s.ParentSpanID.String()
+
+	// Log all Annotations combined with Annotation attributes
+	for _, a := range s.Annotations {
+		aEntry := baseEntry.WithFields(a.Attributes)
+		level := logrus.DebugLevel
+		if _, ok := aEntry.Data[logrus.ErrorKey]; ok {
+			level = logrus.ErrorLevel
+		}
+		aEntry.Time = a.Time
```
kevpar: I don't think we include the Time stored in the Logrus entry in the ETW hook, so I'd recommend logging this as a separate field like `annotationTime`.

jterry75: Why? When this is written out the logrus entry will be "some time after" the annotation actually took place. Why not just have logrus write it out as if it happened at the time the annotation actually occurred?

internal/runtime/hcsv2/uvm.go
```diff
@@ -291,6 +299,6 @@ func (h *Host) ModifyHostSettings(settings *prot.ModifySettingRequest) error {
 
 // Shutdown terminates this UVM. This is a destructive call and will destroy all
 // state that has not been cleaned before calling this function.
-func (h *Host) Shutdown() {
+func (h *Host) Shutdown(ctx context.Context) {
```
jterry75: I will remove this. Its not helpful since the call is destructive

internal/storage/mount_test.go
```diff
@@ -30,7 +31,7 @@ func Test_Unmount_Stat_Valid_Path(t *testing.T) {
 	unixUnmount = func(target string, flags int) error {
 		return nil
 	}
-	err := UnmountPath(expectedName, false)
+	err := UnmountPath(context.TODO(), expectedName, false)
```
jterry75: Oops. These should be context.Background() since its test code its not a `TODO`

service/gcs/bridge/bridge.go
```diff
@@ -126,6 +129,8 @@ func (mux *Mux) ServeMsg(r *Request) (RequestResponse, error) {
 
 // Request is the bridge request that has been sent.
 type Request struct {
+	// Context is the request context received from the bridge.
+	Context context.Context
```
kevpar: Do we send a `context.Context` over the wire? Does that work?

jterry75: We do not. I will extend the protocol to pass trace/span id in the json message but that will come when we have the shim support OC

jterry75:
Only when we call out of the v2 packages. I will work on getting those updated in separate PR's

PR #319, opened 2019-06-17
Maps /etc/hosts and /etc/hostname to sandbox/workload containers

jterry75:
1. Modifies the existing networking model to properly allocate namespaces only
for standalone or sandbox containers.
2. Modifies the existing /etc/resolv.conf handling to properly create for
standalone or sandbox containers but bind mount for workload containers.
3. Adds support for mapping the sandbox containers /etc/hostname to the
workload container.
4. Adds support for mapping the sandbox containers /etc/hosts file to the
workload container.
5. Adds support to standalone containers for /etc/hostname and /etc/hosts
files.

### DIFF
--- a/internal/runtime/hcsv2/network.go
+++ b/internal/runtime/hcsv2/network.go
@@ -1,0 +1,274 @@
+package hcsv2
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/Microsoft/opengcs/internal/network"
+	"github.com/Microsoft/opengcs/service/gcs/gcserr"
+	"github.com/Microsoft/opengcs/service/gcs/prot"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	// namespaceSync protects access to `namespaces`.
+	namespaceSync sync.Mutex
+	// namespaces is the set of `in-memory` namespace adapters know to the GCS.
+	// These may or may not be assigned to a container as there is support for
+	// pre-Add and post-Add.
+	namespaces map[string]*namespace
+
+	networkInstanceIDToName = network.InstanceIDToName
+)
+
+func init() {
+	namespaces = make(map[string]*namespace)
+}
+
+// getNetworkNamespace returns the namespace found by `id`. If the namespace
+// does not exist returns `gcserr.HrErrNotFound`.
+func getNetworkNamespace(id string) (*namespace, error) {
+	id = strings.ToLower(id)
+
+	namespaceSync.Lock()
+	defer namespaceSync.Unlock()
+
+	ns, ok := namespaces[id]
+	if !ok {
+		return nil, gcserr.WrapHresult(errors.Errorf("namespace '%s' not found", id), gcserr.HrErrNotFound)
+	}
+	return ns, nil
+}
+
+// getOrAddNetworkNamespace returns the namespace found by `id` or creates a new
+// one and assigns `id.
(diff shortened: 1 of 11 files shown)


jterry75:
@kevpar - PTAL

internal/runtime/hcsv2/standalone_container.go
```diff
@@ -0,0 +1,139 @@
...
+	}()
+
+	// Generate the standalone root dir
+	rootDir := getStandaloneRootDir(id)
+	if err := os.MkdirAll(rootDir, 0755); err != nil {
+		return errors.Wrapf(err, "failed to create container root directory %q", rootDir)
+	}
+
+	// Write the hostname
+	if !isInMounts("/etc/hostname", spec.Mounts) {
```
kevpar: There are some differences between what we do here and for a sandbox container that I don't understand:

- For sandbox we don't check if the file is in mounts before generating it. Should we do that?
- For standalone, it seems we add another mount for /etc/hostname. Does this mean we'll have two mount entries for it?
- Should we also be adding a mount entry for the sandbox case?

internal/runtime/hcsv2/standalone_container.go
```diff
@@ -0,0 +1,139 @@
...
+			Source:      getStandaloneResolvPath(id),
+			Options:     []string{"bind"},
+		}
+		if isRootReadonly(spec) {
+			mt.Options = append(mt.Options, "ro")
+		}
+		spec.Mounts = append(spec.Mounts, mt)
+	}
+
+	// Clear the windows secion as we dont want to forward to runc
```
kevpar: "section"

internal/runtime/hcsv2/standalone_container.go
```diff
@@ -0,0 +1,139 @@
...
+	if err := os.MkdirAll(rootDir, 0755); err != nil {
+		return errors.Wrapf(err, "failed to create container root directory %q", rootDir)
+	}
+
+	// Write the hostname
+	if !isInMounts("/etc/hostname", spec.Mounts) {
+		hostname := spec.Hostname
+		if hostname == "" {
+			var err error
+			hostname, err = os.Hostname()
```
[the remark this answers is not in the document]
jterry75: This isn't possible. We don't allow multiple standalone containers per UVM

internal/runtime/hcsv2/standalone_container.go
jterry75: - No standalone containers have different semantics. We are forced with the sandbox to create the file every time. Only the workload container is allowed to override. The standalone container is more like the "sandbox + workload" all at once.
- We only have one mount for each of `/etc/hosts`, `/etc/hostname` and `/etc/resolv.conf`
- Sandbox actually has no workload so it needs no mounts. It just for "accounting" to make sure all workloads share these files that they are generated as part of the "sandbox" activation

kevpar: I think I misread `!isInMounts(...)` and missed the `!`. Thus my confusion. :)

internal/runtime/hcsv2/network.go
```diff
@@ -0,0 +1,274 @@
...
+
+	op := "hcsv2::removeNetworkNamespace"
+	log := logrus.WithField("namespace", id)
+	log.Info(op + " - Begin Operation")
+	defer log.Info(op + " - End Operation")
+
+	namespaceSync.Lock()
+	defer namespaceSync.Unlock()
+
+	ns := namespaces[id]
```
kevpar: `ns, ok := namespaces[id]`

kevpar:
Couple comments but overall LGTM

PR #320, opened 2019-06-20
MessageBase should not be including as ptr type

jterry75:


jterry75:
I merged this because its small. Let me know if you see any issues or want something different.

jstarks:
LGTM

PR #328, opened 2019-07-22
Stop logging expected error values

jterry75:
Some GCS* functions return errors to signify a success with additional
information. In these casses we need to skip logging the error because the use
case is expected and this isn't an error for the client who handles it.


jterry75:
Related: https://github.com/microsoft/hcsshim/pull/650

kevpar:
One note: If we end up having a bunch of error filtering like this in the future, I think a helper function like the below would be useful:

```golang
func filterError(err error, filters ...func(err error) bool) error {
    for _, f := range filters {
        if f(err) {
            return nil
        }
    }
    return err
}

oc.SetSpanStatus(span, filterError(err, gcserr.IsNotFound))
```

jterry75:
Closing since we decided the span should return the true operation result.

PR #329, opened 2019-07-25
Send SignalProcess to initpid to all pids in ns

jterry75:

### DIFF
--- a/internal/runtime/hcsv2/process.go
+++ b/internal/runtime/hcsv2/process.go
@@ -114,15 +114,31 @@ func (p *Process) Kill(ctx context.Context, signal syscall.Signal) (err error) {
 		trace.Int64Attribute("pid", int64(p.pid)),
 		trace.Int64Attribute("signal", int64(signal)))
 
-	if err := syscall.Kill(int(p.pid), signal); err != nil {
-		if err == syscall.ESRCH {
-			return gcserr.NewHresultError(gcserr.HrErrNotFound)
-		}
-		return err
-	}
+	// When a container contains more than one process we can fail to unblock
+	// the wait if we only signal the init process. Instead we issue a `runc
+	// kill --all` which then signals all processes in the container.
 	if p.init {
-		p.c.setExitType(signal)
+		if err := p.c.Kill(ctx, signal); err != nil {
+			// Since we use the container logic to kill here it returns a
+			// `gcserr.HrVmcomputeSystemNotFound` but we need a
+			// `gcserr.HrErrNotFound` since we are operating on a process
+			// handle.
+			hr, e2 := gcserr.GetHresult(err)
+			if e2 == nil && hr == gcserr.HrVmcomputeSystemNotFound {
+				return gcserr.NewHresultError(gcserr.HrErrNotFound)
+			}
+			// Unknown runc error, return as is.
+			return err
+		}
+	} else {
+		if err := syscall.Kill(int(p.pid), signal); err != nil {
+			if err == syscall.ESRCH {
+				return gcserr.NewHresultError(gcserr.HrErrNotFound)
+			}
+			return err
+		}
 	}
+
 	return nil
 }
 


internal/runtime/hcsv2/process.go
```diff
@@ -114,15 +114,34 @@ func (p *Process) Kill(ctx context.Context, signal syscall.Signal) (err error) {
...
 	if p.init {
-		p.c.setExitType(signal)
+		if err := p.c.Kill(ctx, signal); err != nil {
+			// Since we use the container logic to kill here it returns a
+			// `gcserr.HrVmcomputeSystemNotFound` but we need a
+			// `gcserr.HrErrNotFound` since we are operating on a process
+			// handle.
+			_, e2 := gcserr.GetHresult(err)
+			if e2 == nil {
+				return gcserr.NewHresultError(gcserr.HrErrNotFound)
```
kevpar: Why don't we return the hr from `GetHresult` here?

internal/runtime/hcsv2/process.go
```diff
@@ -114,15 +114,34 @@ func (p *Process) Kill(ctx context.Context, signal syscall.Signal) (err error) {
...
+			return err
+		}
+	} else {
+		if err := syscall.Kill(int(p.pid), signal); err != nil {
+			if err == syscall.ESRCH {
+				return gcserr.NewHresultError(gcserr.HrErrNotFound)
+			}
+			return err
+		}
+		if p.init {
```
kevpar: It looks like this condition will never occur (since it's inside the `else` block). I think we just want to move the `p.c.setExitType(signal)` to the `if p.init` block?

internal/runtime/hcsv2/process.go
jterry75: Did you read the comment above? I thought that was clear lol must not have been.

internal/runtime/hcsv2/process.go
jterry75: No look at container.Kill it does this already right?

kevpar: Okay but your code here still looks like

```golang
if p.init {
    ...
} else {
    ...
    if p.init {
        ...
    }
}
```

kevpar: Am I crazy? 🤔

internal/runtime/hcsv2/process.go
kevpar: I understand why we need to convert the error into something from `NewHresultError`. I don't understand why we need to convert every HResult into `HrErrNotFound`. We want to return not found even if the kill fails for some other reason?

internal/runtime/hcsv2/process.go
jterry75: Oh I should of looked in the full code view. Yes you are right now that we have the first `if/else` we no longer need this block of code. Will fix.

internal/runtime/hcsv2/process.go
jterry75: The call to Kill ONLY returns `HrVmcomputeSystemNotFound` or it doesn't return a `HRESULT` error. But I can make this explicit by adding a check for that I guess.

internal/runtime/hcsv2/process.go
```diff
@@ -114,15 +114,31 @@ func (p *Process) Kill(ctx context.Context, signal syscall.Signal) (err error) {
...
+	// kill --all` which then signals all processes in the container.
 	if p.init {
-		p.c.setExitType(signal)
+		if err := p.c.Kill(ctx, signal); err != nil {
+			// Since we use the container logic to kill here it returns a
+			// `gcserr.HrVmcomputeSystemNotFound` but we need a
+			// `gcserr.HrErrNotFound` since we are operating on a process
+			// handle.
+			hr, e2 := gcserr.GetHresult(err)
+			if e2 != nil && hr == gcserr.HrVmcomputeSystemNotFound {
```
kevpar: I think this should be `e2 == nil` right?

jterry75: You are 100% correct. At least one of us is paying attention. Bugs in making the changes you wanted lol. Worked before my bad.

jterry75:
@kevpar - 4th times the charm?

kevpar:
LGTM!

PR #332, opened 2019-07-31
Use context wait patterns

jterry75:

### DIFF
--- a/internal/runtime/hcsv2/uvm.go
+++ b/internal/runtime/hcsv2/uvm.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/Microsoft/opengcs/internal/oc"
 	"github.com/Microsoft/opengcs/internal/storage"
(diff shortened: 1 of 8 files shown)


internal/storage/scsi/scsi.go
```diff
@@ -96,22 +94,24 @@ func ControllerLunToName(ctx context.Context, controller, lun uint8) (_ string,
...
+	// under /sys/bus/scsi/devices/<scsiID>/block.
+	blockPath := filepath.Join("/sys/bus/scsi/devices", scsiID, "block")
 	var deviceNames []os.FileInfo
-	startTime := time.Now()
 	for {
-		// Devices matching the given SCSI code should each have a subdirectory
-		// under /sys/bus/scsi/devices/<scsiID>/block.
-		var err error
-		deviceNames, err = ioutil.ReadDir(filepath.Join("/sys/bus/scsi/devices", scsiID, "block"))
+		deviceNames, err = ioutil.ReadDir(blockPath)
```
kevpar: It looks like if this call succeeds we never exit the loop.

jterry75: I have no idea how this passed tests. Good catch. Let me fix and try again.

internal/runtime/hcsv2/uvm.go
```diff
@@ -204,8 +205,10 @@ func newInvalidRequestTypeError(rt prot.ModifyRequestType) error {
 func modifyMappedVirtualDisk(ctx context.Context, rt prot.ModifyRequestType, mvd *prot.MappedVirtualDiskV2) (err error) {
 	switch rt {
 	case prot.MreqtAdd:
+		mountCtx, cancel := context.WithTimeout(ctx, time.Second*4)
```
kevpar: FYI, looks like this timeout is changing from 2 to 4 seconds. Not sure if that's intentional.

jterry75: There were two calls that each had 2 second timeouts. Now the entire set (both calls) have up to the total 4 seconds.

PR #333, opened 2019-07-31
Break apart hcsv2.Host ModifySettings methods

jterry75:
Seperates all hvsv2.Host ModifySettings calls into seperate functions to make
growth easier when adding new entries.


jterry75:
@kevpar - No changes just broke it apart.

PR #335, opened 2019-08-14
Add ability to dump stacks in OpenGCS

katiewasnothere:
Added bridge handler to dump opengcs stacks

### DIFF
--- a/internal/debug/dumpstacks.go
+++ b/internal/debug/dumpstacks.go
@@ -1,0 +1,15 @@
+package debug
+
+import "runtime"
+
+func DumpStacks() string {
+	buf := make([]byte, 4096)
+	for {
+		buf = buf[:runtime.Stack(buf, true)]
+		if len(buf) < cap(buf) {
+			break
+		}
+		buf = make([]byte, 2*len(buf))
+	}
+	return string(buf)
+}
--- a/service/gcs/main.go
+++ b/service/gcs/main.go
@@ -5,10 +5,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"os/signal"
-	"path/filepath"
-	"runtime"
-	"sync"
 	"time"
 
 	"github.com/Microsoft/opengcs/internal/oc"
@@ -19,7 +15,6 @@ import (
 	"github.com/Microsoft/opengcs/service/gcs/transport"
 	"github.com/sirupsen/logrus"
 	"go.opencensus.io/trace"
-	"golang.org/x/sys/unix"
 )
 
 func main() {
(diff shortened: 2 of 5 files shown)


katiewasnothere:
@jterry75  do we still want the ability to write to a file in dump stacks? main.go uses the write to file ability, but @kevpar and I are unsure if this is needed.

service/gcs/prot/protocol.go
```diff
@@ -99,6 +99,8 @@ const (
 	ComputeSystemModifySettingsV1 = 0x10100a01
 	// ComputeSystemNegotiateProtocolV1 is the protocol negotiation request.
 	ComputeSystemNegotiateProtocolV1 = 0x10100b01
+	// ComputeSystemDumpStackV1 is the dump stack request
+	ComputeSystemDumpStacksV1 = 0x10100c01
```
jterry75: You will need a Response identifier as well.

service/libs/commonutils/utilities.go
```diff
@@ -24,3 +28,41 @@ func DecodeJSONWithHresult(r io.Reader, v interface{}) error {
 	}
 	return nil
 }
+
+func DumpStacks(baseLogPath string, writeToFile bool) {
```
jterry75: Since this is a guest and its different than the host. We don't need `writeToFile` support and we should return this as part of the RPC rather than using `stdout` propagation.

service/gcs/main.go
```diff
@@ -17,6 +14,7 @@ import (
 	"github.com/Microsoft/opengcs/service/gcs/core/gcs"
 	"github.com/Microsoft/opengcs/service/gcs/runtime/runc"
 	"github.com/Microsoft/opengcs/service/gcs/transport"
+	"github.com/Microsoft/opengcs/service/libs/commonutils"
```
jterry75: I know the code has these but this is *legacy*. Golang should use `internal/<pkg>` or `pkg/<pkg>`. Since this is internal lets call it `internal/debug` or something.

internal/debug/dumpstacks.go
```diff
@@ -0,0 +1,19 @@
...
+import "runtime"
+
+func DumpStacks() string {
+	var (
+		buf       []byte
+		stackSize int
+	)
+	bufStartLen := 10240
+
+	for stackSize == len(buf) {
```
jterry75: Does this actually work? The logic seems highly suspect. I would do:

https://github.com/microsoft/hcsshim/blob/0291ac21db715cc43d6068dc18f78b7abed6ab9d/cmd/containerd-shim-runhcs-v1/service.go#L359

PR #338, opened 2019-08-30
Implement the stub methods for LVM modification

jterry75:

### DIFF
--- a/service/gcs/prot/protocol.go
+++ b/service/gcs/prot/protocol.go
@@ -266,6 +266,7 @@ type GcsGuestCapabilities struct {
 	NamespaceAddRequestSupported bool `json:",omitempty"`
 	SignalProcessSupported       bool `json:",omitempty"`
 	DumpStacksSupported          bool `json:",omitempty"`
+	LogicalVolumeSupported       bool `json:",omitempty"`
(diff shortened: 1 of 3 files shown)


service/gcs/prot/protocol.go
```diff
@@ -932,3 +940,11 @@ type PropertyQuery struct {
 type Properties struct {
 	ProcessList []ProcessDetails `json:",omitempty"`
 }
+
+// LogicalVolume represents a logical volume that is created on top of a set of
+// physical devices.
+type LogicalVolume struct {
+	ID        string
+	MountPath string
+	Devices   []MappedVirtualDiskV2
```
kevpar: I notice `MappedVirtualDiskV2` itself also has a `MountPath`. Do we use that field for something in the LVM case, or should we maybe make a new type that's just the disk identifiers without a mount path, to use here?

jterry75: I'm not sure I care. I can make a new type if you want

jterry75: But I can see how its confusing at best to reuse this type with the precondition that you dont set the MountPoint in the disk itself

PR #339, opened 2019-09-04
init: Add option to initialize entropy from vsock

jstarks:
This change adds a command line option, -e, that specifies a vsock port
to read initial RNG entropy. init reads all data written to the
connection by the host, adds this data to the kernel RNG, and increments
the available entropy count. The host is trusted to write
cryptographically random data to this port.

Without this, the available entropy at LCOW boot is very low, and data
read from /dev/urandom is likely to be highly predicable.

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -2,6 +2,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <getopt.h>
+#include <linux/random.h>
 #include <net/if.h>
 #include <netinet/ip.h>
 #include <signal.h>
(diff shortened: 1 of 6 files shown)


init/init.c
```diff
@@ -236,6 +238,42 @@ void init_network(const char *iface, int domain) {
...
+    }
+
+    struct {
+        int entropy_count;
+        int buf_size;
+        char buf[4096];
+    } buf;
+
+    for (;;) {
+        ssize_t n = read(s, buf.buf, sizeof(buf.buf));
```
jterry75: So I assume the host will then send/close which causes the `n == 0` break below and this returns?

jstarks: Right, when host closes we get n == 0. This loop would be a little more complicated in the presence of signals, but we disable signals early on (and I'm not sure where those signals would come from anyway).

jstarks:
Ready to merge.

PR #347, opened 2019-10-02
Add sock options and set payload size for plan 9 mounts

katiewasnothere:
This is a performance improvement for 9p mounts that takes advantage of larger socket bandwidth and message size. 

### DIFF
--- a/internal/storage/plan9/plan9.go
+++ b/internal/storage/plan9/plan9.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"syscall"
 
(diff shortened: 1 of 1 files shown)


internal/storage/plan9/plan9.go
```diff
@@ -14,6 +15,8 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+const PacketPayloadBytes = 65536
```
jterry75: NIT: can we not export this since its not used outside the package?

katiewasnothere: Fixed.

PR #349, opened 2019-10-04
Support mount propagation for sandbox mounts

katiewasnothere:
* This change creates a new dedicated directory path in the uvm for all
sandbox mounts at /tmp/gcs/cri/<sandboxid>/sandboxMounts.
* The above path will be marked as rshared
* OpenGCS now expects sandbox mounts to be given with a source prefix
of 'sandbox://'
* OpenGCS will also create any files or directories in the resulting
source sandbox mount path.

Needs: https://github.com/microsoft/hcsshim/pull/710

### DIFF
--- a/internal/runtime/hcsv2/workload_container.go
+++ b/internal/runtime/hcsv2/workload_container.go
@@ -2,7 +2,9 @@ package hcsv2
 
 import (
 	"context"
+	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/Microsoft/opengcs/internal/log"
 	"github.com/Microsoft/opengcs/internal/oc"
@@ -16,6 +18,31 @@ func getWorkloadRootDir(sbid, id string) string {
 	return filepath.Join(getSandboxRootDir(sbid), id)
 }
 
+func updateSandboxMounts(sbid string, spec *oci.Spec) error {
+	sandboxMountPrefix := "sandbox://"
+	for i, m := range spec.Mounts {
+		if strings.HasPrefix(m.Source, sandboxMountPrefix) {
+			mountsDir := getSandboxMountsDir(sbid)
+			subPath := strings.TrimPrefix(m.Source, sandboxMountPrefix)
+			sandboxSource := filepath.Join(mountsDir, subPath)
+
+			if !strings.HasPrefix(sandboxSource, mountsDir) {
+				return errors.Errorf("mount path %v for mount %v is not within sandbox's mounts dir", sandboxSource, m.Source)
+			}
+
+			spec.Mounts[i].Source = sandboxSource
+
+			_, err := os.Stat(sandboxSource)
+			if os.IsNotExist(err) {
+				if err := os.MkdirAll(sandboxSource, 0755); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
 func setupWorkloadContainerSpec(ctx context.Context, sbid, id string, spec *oci.Spec) (err error) {
 	ctx, span := trace.StartSpan(ctx, "hcsv2::setupWorkloadContainerSpec")
 	defer span.End()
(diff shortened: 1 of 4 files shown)


internal/runtime/hcsv2/workload_container.go
```diff
@@ -16,6 +18,40 @@ func getWorkloadRootDir(sbid, id string) string {
...
+func updateSandboxMounts(sbid string, spec *oci.Spec) error {
+	sandboxMountPrefix := "sandbox://"
+	for i, m := range spec.Mounts {
+		if strings.HasPrefix(m.Source, sandboxMountPrefix) {
+			subPath := strings.TrimPrefix(m.Source, sandboxMountPrefix)
+			sandboxSource := filepath.Join(getSandboxMountsDir(sbid), subPath)
+			spec.Mounts[i].Source = sandboxSource
+
+			_, err := os.Stat(sandboxSource)
+			if os.IsNotExist(err) {
```
jterry75: I dont think we should handle creating a file. If we do a `os.Stat` on sandbox source we should just issue a mkdir all and move on. If the stat comes back as valid then its either a file or directory within the sandbox mount and is already a valid source.

internal/runtime/hcsv2/workload_container.go
```diff
@@ -16,6 +18,40 @@ func getWorkloadRootDir(sbid, id string) string {
...
+	_, err := os.Create(sandboxSource)
+	return err
+}
+
+func updateSandboxMounts(sbid string, spec *oci.Spec) error {
+	sandboxMountPrefix := "sandbox://"
+	for i, m := range spec.Mounts {
+		if strings.HasPrefix(m.Source, sandboxMountPrefix) {
+			subPath := strings.TrimPrefix(m.Source, sandboxMountPrefix)
+			sandboxSource := filepath.Join(getSandboxMountsDir(sbid), subPath)
```
jterry75: We need to verify that sandboxSource is within `getSandboxMountsDir` after this is joined.

katiewasnothere: What is the concern here?

jterry75: I'm not sure what you mean. We need to verify that the resulting `sandboxSource` is still under `getSandboxMountsDir(id)` after its joined. Consider a path like `..\..\tmp` or something. We dont want people to be able to do that.

katiewasnothere: Makes sense

katiewasnothere: Done

internal/runtime/hcsv2/workload_container.go
katiewasnothere: Done

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.12
 
 require (
 	github.com/Microsoft/go-winio v0.4.12
-	github.com/Microsoft/hcsshim v0.0.0-20190404205635-cf1c2137fae1
+	github.com/Microsoft/hcsshim v0.8.6
```
jterry75: Why this? This is wrong since the commit is much newer than the relase

katiewasnothere: Sorry, updated!

jterry75: I'm not against this change I'm just wondering why we are doing it? Was this needed for these changes?

katiewasnothere: I don't think so, was just updating since it appeared to have a fairly old commit for the shim. I can revert if needed.

jterry75: Its ok. In the future lets try to make these as separate commits because it makes it easier to find regressions.

jterry75:
@katiewasnothere - Just needs a rebase it looks like and we are good.

go.mod
katiewasnothere: Will do!

go.mod
```diff
@@ -3,10 +3,9 @@ module github.com/Microsoft/opengcs
 go 1.12
 
 require (
-	github.com/Microsoft/go-winio v0.4.12
-	github.com/Microsoft/hcsshim v0.0.0-20190404205635-cf1c2137fae1
-	github.com/containerd/containerd v1.3.0
-	github.com/containerd/continuity v0.0.0-20190815185530-f2a389ac0a02 // indirect
+	github.com/Microsoft/go-winio v0.4.15-0.20190919025122-fc70bd9a86b5
+	github.com/Microsoft/hcsshim v0.8.7-0.20191007171158-cf76d8132ece
+	github.com/containerd/containerd v1.3.0-beta.2.0.20190828155532-0293cbd26c69
```
jterry75: Sorry this one is older than v1.3.0. How are you doing this it seems odd the changes its making here.

katiewasnothere: By running go get github.com/Microsoft/hcsshim@master.

katiewasnothere: By running go get github.com/Microsoft/hcsshim@master.

PR #350, opened 2019-10-10
Create Containers and GCS cgroups on GCS startup

jterry75:
Setup the UVM cgroups to protect against a workload taking all available
memory and causing the GCS to malfunction we create two cgroups: gcs,
containers.

The containers cgroup is limited only by {Totalram - 75 MB (reservation)}.

The gcs cgroup is limited to 10 MB to prevent unknown memory leaks over time
from affecting workload memory.


internal/runtime/hcsv2/sandbox_container.go
```diff
@@ -97,6 +97,9 @@ func setupSandboxContainerSpec(ctx context.Context, id string, spec *oci.Spec) (
 	// also has a concept of a sandbox/shm file when the IPC NamespaceMode !=
 	// NODE.
 
+	// Force the parent cgroup into our /containers root
+	spec.Linux.CgroupsPath = "/containers/" + id
```
jterry75: @katiewasnothere - For the cgroup stats work you are doing with this change you can now predict the path very easily for each subsystem. The root is "/containers/{id}" under each root. So:
```
/sys/fs/cgroup/cpu/containers/{id}
/sys/fs/cgroup/memory/containers/{id}
etc.
```

service/gcs/main.go
```diff
@@ -110,6 +111,43 @@ func main() {
 		bridgeOut = bridgeCon
 	}
 
+	// Setup the UVM cgroups to protect against a workload taking all available
```
jterry75: @jstarks - Mind double checking the idea here. Thanks

jstarks: Seems like a reasonable start. Maybe add some command line parameters to configure the sizes, though, so that we can adjust this from the host.

PR #354, opened 2019-10-21
output log info from failures to run runc commands for create/exec

katiewasnothere:

### DIFF
--- a/service/gcs/runtime/runc/runc.go
+++ b/service/gcs/runtime/runc/runc.go
@@ -546,6 +546,7 @@ func (c *container) startProcess(tempProcessDir string, hasTerminal bool, stdioS
 	}
 
 	logPath := c.r.getLogPath(c.id)
+	args = append([]string{"--log-format", "json"}, args...)
 	args = append([]string{"--log", logPath}, args...)
 
 	args = append(args, "--pid-file", filepath.Join(tempProcessDir, "pid"))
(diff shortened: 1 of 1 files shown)


service/gcs/runtime/runc/runc.go
```diff
@@ -587,7 +587,11 @@ func (c *container) startProcess(tempProcessDir string, hasTerminal bool, stdioS
 	}
 
 	if err := cmd.Run(); err != nil {
-		return nil, errors.Wrapf(err, "failed to run runc create/exec call for container %s", c.id)
+		b, readErr := ioutil.ReadFile(logPath)
```
katiewasnothere: @jterry75 Can you provide feedback on this? The only way I could find to get the actual error logged to stderr from running the cmd was by reading the runc log that's in the pod. This log file includes info like log level and timestamp of the message. What are you thoughts on simply logging in opengcs the runc file from the pod instead of returning it in the error to startProcess?

katiewasnothere: Updated code with suggestions from John.

service/gcs/runtime/runc/runc.go
```diff
@@ -587,7 +588,15 @@ func (c *container) startProcess(tempProcessDir string, hasTerminal bool, stdioS
...
+		if readErr != nil {
+			return nil, errors.Wrapf(readErr, "failed to read runc log for runc create/exec call failure for container %s", c.id)
+		}
+		var result map[string]interface{}
```
jterry75: Isnt this just a logrus entry array?

jterry75: I mean, can it be more than one?

jterry75: Oh I see. There could be more than just an error in here. I wonder if we want to do some wrapping up to the last error for the extended info?

katiewasnothere:
Closing in favor of #355

PR #355, opened 2019-10-29
add ability to read runc log for errors in runc commands

katiewasnothere:
This work allows better error response for users when commands to runc fail. 

### DIFF
--- a/service/gcs/runtime/runc/runc.go
+++ b/service/gcs/runtime/runc/runc.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
-	"os/exec"
 	"path"
 	"path/filepath"
 	"strconv"
@@ -122,11 +121,13 @@ func (r *runcRuntime) CreateContainer(id string, bundlePath string, stdioSet *st
 // CreateContainer.
 func (c *container) Start() error {
 	logPath := c.r.getLogPath(c.id)
-	cmd := exec.Command("runc", "--log", logPath, "start", c.id)
+	args := []string{"start", c.id}
+	cmd := createRuncCommand(logPath, args...)
 	out, err := cmd.CombinedOutput()
 	if err != nil {
+		runcErr := getRuncLogError(logPath)
 		c.r.cleanupContainer(c.id)
-		return errors.Wrapf(err, "runc start failed with: %s", out)
+		return errors.Wrapf(err, "runc start failed with %v: %s", runcErr, string(out))
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


service/gcs/runtime/runc/runc.go
```diff
@@ -122,11 +122,12 @@ func (r *runcRuntime) CreateContainer(id string, bundlePath string, stdioSet *st
 // CreateContainer.
 func (c *container) Start() error {
 	logPath := c.r.getLogPath(c.id)
-	cmd := exec.Command("runc", "--log", logPath, "start", c.id)
+	cmd := exec.Command("runc", "--log", logPath, "--log-format", "json", "start", c.id)
```
jstarks: Would be nice to make a helper for this (`exec.Command("runc", "--log", logPath, "--log-format", "json", args...)`)

katiewasnothere: Done. Let me know if my change is not what you had in mind.

jterry75:
@katiewasnothere - I am still concerned (can you tell me if I shouldn't be) that this can be an array. And since we use the same file over time it seems that the "last error" would need to be read from the back of the file?

katiewasnothere:
@jterry75 I updated to return the last error in the file. Since this code path is only hit in the event of an error, I don't think truly reading from the back of the file is necessary. If you disagree, I can change it.

service/gcs/runtime/runc/utils.go
```diff
@@ -88,3 +92,40 @@ func (r *runcRuntime) processExists(pid int) bool {
...
+	reader, err := os.OpenFile(logPath, syscall.O_RDONLY, 0644)
+	if err != nil {
+		return nil
+	}
+	defer reader.Close()
+
+	var lastErr interface{}
+	dec := json.NewDecoder(reader)
+	for {
+		var out map[string]interface{}
```
jterry75: Ok last thing, a lot of time code looks like this `logrus.WithError(err).Error("we hit an error")`. This code would return "we hit an error" but we should also capture `err` as well.

jstarks: It's interesting to me that we're unmarshaling the JSON into a map instead of into a struct like we did in hcsshim when handling the same case... Is there a reason for this?

katiewasnothere: The json decoder used here doesn't differentiate between an error that is the result of the stream being broken and an error from unmarshalling. However, I'm not sure that matters too much since in the event the logs aren't returned to the user on error, we can still manually check the logs in the uvm. I'll update it.

katiewasnothere: Done.

PR #358, opened 2019-11-06
Remove v2 dependency on v1 for host processes

jterry75:
When activating host processes (ie: processes in the uvm namespace) the
v2 code still used the old v1 execution. This change removes that
dependency in preparation on removing v1 entirely.


jterry75:
@katiewasnothere - Part 1 of removing v1 to make this change more clear.

jterry75:
@katiewasnothere - Ok verified this works as expected.

PR #360, opened 2019-11-08
add new delete container state bridge call

katiewasnothere:
* add delete container state capability for opengcs
* delete container state bridge call calls runc delete for container
* delete container state bridge call unmounts user mounts on sandbox
mounts
* add new function to unmount all paths under a given path
* add tests for new unmount function

### DIFF
--- a/internal/runtime/hcsv2/container.go
+++ b/internal/runtime/hcsv2/container.go
@@ -7,6 +7,8 @@ import (
 	"sync"
 	"syscall"
 
+	"github.com/Microsoft/opengcs/internal/log"
+	"github.com/Microsoft/opengcs/internal/storage"
 	"github.com/Microsoft/opengcs/service/gcs/gcserr"
 	"github.com/Microsoft/opengcs/service/gcs/prot"
 	"github.com/Microsoft/opengcs/service/gcs/runtime"
@@ -23,7 +25,8 @@ type Container struct {
 	id    string
 	vsock transport.Transport
 
-	spec *oci.Spec
+	spec      *oci.Spec
+	isSandbox bool
 
 	container   runtime.Container
 	initProcess *containerProcess
@@ -114,6 +117,16 @@ func (c *Container) Kill(ctx context.Context, signal syscall.Signal) error {
 	}
 	c.setExitType(signal)
 	return nil
+}
+
+func (c *Container) Delete(ctx context.Context) error {
+	if c.isSandbox {
+		// remove user mounts in sandbox container
+		if err := storage.UnmountAllInPath(ctx, getSandboxMountsDir(c.id), true); err != nil {
+			log.G(ctx).WithError(err).Error("failed to unmount sandbox mounts")
+		}
+	}
+	return c.container.Delete()
 }
 
 // Wait waits for the container's init process to exit.
--- a/internal/storage/mount.go
+++ b/internal/storage/mount.go
@@ -3,9 +3,11 @@
 package storage
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"syscall"
 
 	"github.com/Microsoft/opengcs/internal/oc"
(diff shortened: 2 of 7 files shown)


internal/storage/mount.go
```diff
@@ -67,3 +75,41 @@ func UnmountPath(ctx context.Context, target string, removeTarget bool) (err err
...
+	for i := len(childMounts) - 1; i >= 0; i-- {
+		childPath := childMounts[i]
+		if err := UnmountPath(ctx, childPath, removeTarget); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func ListMountPointsUnderPath(path string) ([]string, error) {
```
jterry75: We shouldn't export this since nobody uses it outside this package.

internal/storage/mount.go
```diff
@@ -67,3 +75,41 @@ func UnmountPath(ctx context.Context, target string, removeTarget bool) (err err
...
+		if len(fields) < numProcMountFields {
+			continue
+		}
+		destPath := fields[1]
+		if strings.HasPrefix(destPath, path) {
+			mountPoints = append(mountPoints, destPath)
+		}
+	}
+
+	return mountPoints, nil
```
jterry75: I think you will need to apply some sort order. Is there some guarantee that `/proc/self/mountinfo` is ordered in a way that we will always create the array in an unmount order?

internal/storage/mount.go
```diff
@@ -14,12 +16,18 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+var (
```
jterry75: `const`

service/gcs/bridge/bridge_v2.go
```diff
@@ -495,3 +496,28 @@ func (b *Bridge) dumpStacksV2(r *Request) (_ RequestResponse, err error) {
 		GuestStacks: stacks,
 	}, nil
 }
+
+func (b *Bridge) deleteContainerStateV2(r *Request) (_ RequestResponse, err error) {
+	ctx, span := trace.StartSpan(r.Context, "opengcs::bridge::deleteContainerStateV2")
+	defer span.End()
+	defer func() { oc.SetSpanStatus(span, err) }()
+
+	span.AddAttributes(
```
jterry75: You can do this in one line

internal/runtime/hcsv2/container.go
```diff
@@ -116,6 +118,16 @@ func (c *Container) Kill(ctx context.Context, signal syscall.Signal) error {
 	return nil
 }
 
+func (c *Container) DeleteContainerState(ctx context.Context) error {
```
jterry75: Lets just call this `Delete`

internal/runtime/hcsv2/container.go
```diff
@@ -116,6 +118,16 @@ func (c *Container) Kill(ctx context.Context, signal syscall.Signal) error {
...
+	if c.isSandbox {
+		// remove user mounts in sandbox container
+		if err := storage.UnmountAllInPath(ctx, getSandboxMountsDir(c.id), true); err != nil {
+			return err
+		}
+	}
+	return c.container.Delete()
```
jterry75: You will also need to remove this from the UVM containers map

internal/runtime/hcsv2/container.go
```diff
@@ -116,6 +118,16 @@ func (c *Container) Kill(ctx context.Context, signal syscall.Signal) error {
 	return nil
 }
 
+func (c *Container) DeleteContainerState(ctx context.Context) error {
+	if c.isSandbox {
+		// remove user mounts in sandbox container
+		if err := storage.UnmountAllInPath(ctx, getSandboxMountsDir(c.id), true); err != nil {
```
jterry75: Should this be best effort and just log the error? IE: If unmount fails, we should probably still tear down the container right?

internal/storage/mount.go
katiewasnothere: I believe /proc/mounts should be listed in the order that the mounts were issued.

internal/runtime/hcsv2/container.go
katiewasnothere: Are you okay with leaving the bridge call as "deleteContainerState"?

internal/storage/mount.go
jterry75: Ok as long as that is true then this will always work in reverse order.

internal/runtime/hcsv2/container.go
jterry75: Lets call it `ComputeSystemDeleteV1`

jterry75: That matches the others better

jterry75: But really we dont need this func. We can do this automatically when the publish notification is done. Maybe we should just move to the ttrpc protocol for LCOW?

PR #364, opened 2019-12-16
Add support for GCS reserve memory

jterry75:

### DIFF
--- a/service/gcs/main.go
+++ b/service/gcs/main.go
@@ -26,7 +26,8 @@ func main() {
 	logFormat := flag.String("log-format", "text", "Logging Format: text or json")
 	useInOutErr := flag.Bool("use-inouterr", false, "If true use stdin/stdout for bridge communication and stderr for logging")
 	v4 := flag.Bool("v4", false, "enable the v4 protocol support and v2 schema")
-	rootMemReserveBytes := flag.Uint64("root-mem-reserve-bytes", 75*1024*1024, "the amount of memory reserved for the orchestration, the rest will be assigned to containers")
+	rootMemReserveBytes := flag.Uint64("root-mem-reserve-bytes", 60*1024*1024, "the amount of memory reserved for the root, the rest will be assigned to containers")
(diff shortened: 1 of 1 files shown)


kevpar:
Should we default the gcs limit to 0 (no limit) instead? I'm not sure what the limit accomplishes since if it's hit, the whole pod will go down anyways, and is currently hard to troubleshoot in that state. We could leave in the ability to configure a limit via command line still.

service/gcs/main.go
```diff
@@ -126,7 +131,7 @@ func main() {
 	if err := syscall.Sysinfo(&sinfo); err != nil {
```
mkatri: should the 75MB number in the comment above be increased to 110MB?

mkatri:
> Should we default the gcs limit to 0 (no limit) instead? I'm not sure what the limit accomplishes since if it's hit, the whole pod will go down anyways, and is currently hard to troubleshoot in that state. We could leave in the ability to configure a limit via command line still.

Wouldn't we at least know which container misbehaved? GCS or workload?

service/gcs/main.go
jterry75: Good catch the comment is now wrong.

jterry75:
> Should we default the gcs limit to 0 (no limit) instead? I'm not sure what the limit accomplishes since if it's hit, the whole pod will go down anyways, and is currently hard to troubleshoot in that state. We could leave in the ability to configure a limit via command line still.

I think as a best practice we should keep the gcs cgroup and use its limit. I agree that it will crash if it hits this limit but it would be nice to know. There is no reason that we should be seeing 50mb of usage and if we do keep seeing crashes we know this is not the real issue. Without the limit there is no way to know if a gcs is running with 200MB and stealing from workload memory.

jterry75:
**DO NOT MERGE** doing some studies to try and find the right numbers here.

kevpar:
Closing this since we took another fix for this issue: https://github.com/microsoft/opengcs/pull/372

PR #367, opened 2020-01-13
[WIP] Continuously log /dev/kmsg via goroutine

kevpar:
Main concern before merging is to test that we aren't going to increase log volume by a huge amount.

We will now spin up a goroutine when the GCS starts which reads log
entries from /dev/kmsg and logs them via Logrus.

Signed-off-by: Kevin Parsons <kevpar@microsoft.com>

### DIFF
--- a/internal/kmsg/kmsg.go
+++ b/internal/kmsg/kmsg.go
@@ -1,0 +1,63 @@
+// Package kmsg contains support for parsing Linux kernel log entries read from
+// /dev/kmsg. These are the same log entries that can be read via the `dmesg`
+// command. Each read from /dev/kmsg is guaranteed to return a single log entry,
+// so no line-splitting is required. If the read buffer is not large enough to
+// contain the log entry, `os.ErrInvalid` is returned.
+//
+// More information can be found here:
+// https://www.kernel.org/doc/Documentation/ABI/testing/dev-kmsg
+package kmsg
+
+import (
+	"errors"
+	"strconv"
+	"strings"
+)
+
+// Entry is a single log entry in kmsg.
+type Entry struct {
+	SyslogPriority     uint8
+	SyslogFacility     uint8
+	Seq                uint64
+	TimeSinceBootMicro uint64
+	Flags              string
+	Message            string
+}
+
+var (
+	// ErrInvalidFormat indicates the kmsg entry failed to parse.
+	ErrInvalidFormat = errors.New("invalid kmsg format")
+)
+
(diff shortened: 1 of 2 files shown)


service/gcs/main.go
```diff
@@ -20,6 +21,45 @@ import (
...
+	if err != nil {
+		logrus.WithError(err).Error("failed to open kmsg")
+		return
+	}
+	// /dev/kmsg never returns EOF, so we just sit in the loop waiting for more
+	// log entries forever.
+	for {
+		// Buffer size from CONSOLE_EXT_LOG_MAX	in Linux's printk.h.
+		b := [8192]byte{}
+		n, err := f.Read(b[:])
```
katiewasnothere: why is the b[:] necessary? can you not just pass through b here? Was this a style choice?

service/gcs/main.go
```diff
@@ -20,6 +21,45 @@ import (
...
+		if err != nil {
+			logrus.WithError(err).Error("kmsg read failure")
+		} else {
+			line := string(b[:n])
+			ke, err := kmsg.Parse(line)
```
katiewasnothere: is it guaranteed that we will only read one line at a time if we're using a buffer with 8192 bytes?

internal/kmsg/kmsg.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// Parse takes a single kmsg log entry string and returns a struct representing
+// the components of the log entry.
+func Parse(s string) (*Entry, error) {
+	fields := strings.SplitN(s, ";", 2)
+	if len(fields) < 2 {
+		return nil, ErrInvalidFormat
+	}
+	prefixFields := strings.SplitN(fields[0], ",", 5)
```
katiewasnothere: can you explain what you're doing here/how a line is expected to be formatted like?

katiewasnothere: Update: this is explained in the above comment which links to [this](https://www.kernel.org/doc/Documentation/ABI/testing/dev-kmsg}

service/gcs/main.go
katiewasnothere: Talked to Kevin offline, this converts an array into a slice.

service/gcs/main.go
katiewasnothere: Again, this is in the comment at the top.

dcantah:
Can close now that this has been merged https://github.com/microsoft/opengcs/pull/373

PR #368, opened 2020-01-30
init: Set RLIMIT_NOFILE hard limit to 1M

jstarks:
Many containers expect an rlimit hard limit greater than the kernel
default of 4096. On Debian systems, the default hard limit is 1M.
In 2018, systemd set the default hard limit to 512K for all processes it
launces. Apparently modern kernel versions can handle large limits
without disadvantage: large limits are efficient, and large fd memory use
is correctly accounted in cgroups.

Therefore, unconditionally set a large hard limit. Keep the soft
limit small for compatibility (an unprivileged process can update
the soft limit up to the hard limit).

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -9,6 +9,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mount.h>
+#include <sys/resource.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


jstarks:
Note that I have not actually tested this other than to ensure it builds. :)

kevpar:
Tested :)

PR #369, opened 2020-01-30
Guest GPU assigned device support

katiewasnothere:
This work sets up the execution of [nvidia-container-cli](https://github.com/NVIDIA/libnvidia-container) as a prestart hook to runc.
* parse container spec options for the gpu device
* build the nvidia-container-cli tool arguments 
* build prestart hook for running nvidia tool
* create new prestart hook binary 
 
Related to https://github.com/microsoft/hcsshim/pull/765

### DIFF
--- a/internal/network/network.go
+++ b/internal/network/network.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/Microsoft/opengcs/internal/log"
 	"github.com/Microsoft/opengcs/internal/oc"
+	"github.com/Microsoft/opengcs/internal/storage/vmbus"
 	"github.com/pkg/errors"
 	"go.opencensus.io/trace"
 )
@@ -111,7 +112,9 @@ func InstanceIDToName(ctx context.Context, id string) (_ string, err error) {
 	id = strings.ToLower(id)
 	span.AddAttributes(trace.StringAttribute("adapterInstanceID", id))
 
-	devicePath := filepath.Join("/sys", "bus", "vmbus", "devices", id, "net")
+	vmBusSubPath := filepath.Join(id, "net")
+	devicePath, err := vmbus.WaitForDevicePath(ctx, vmBusSubPath)
+
 	var deviceDirs []os.FileInfo
 	for {
 		deviceDirs, err = ioutil.ReadDir(devicePath)
--- a/internal/runtime/hcsv2/nvidia_utils.go
+++ b/internal/runtime/hcsv2/nvidia_utils.go
@@ -1,0 +1,91 @@
+// +build linux
+
+package hcsv2
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/Microsoft/opengcs/internal/storage/pci"
+	oci "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+)
+
+// path that the shim mounts the nvidia gpu vhd to in the uvm
+// this MUST match the path mapped to in the shim
+const lcowNvidiaMountPath = "/run/nvidia"
+
+// annotation to find the gpu capabilities on the container spec
+// must match the hcsshim annotation string for gpu capabilities
+const annotationContainerGPUCapabilities = "io.microsoft.container.gpu.capabilities"
+const nvidiaDebugFilePath = "/nvidia-container.log"
+
+// TODO katiewasnothere: prestart hooks will be depracated, this needs to be moved to a createRuntime hook
+// described here: https://github.com/opencontainers/runtime-spec/blob/39c287c415bf86fb5b7506528d471db5405f8ca8/config.md#posix-platform-hooks
+// addNvidiaDevicePreHook builds the arguments for nvidia-container-cli and creates the prestart hook
+func addNvidiaDevicePreHook(ctx context.Context, spec *oci.Spec) error {
+	nvidiaToolBinary := "nvidiaPrestartHook"
+	nvidiaToolPath, err := exec.LookPath(nvidiaToolBinary)
+	if err != nil {
+		return errors.Wrapf(err, "failed to find %s for container GPU support", nvidiaToolBinary)
+	}
+
+	debugOption := fmt.Sprintf("--debug=%s", nvidiaDebugFilePath)
+
+	// TODO katiewasnothere: right now both host and container ldconfig do not work as expected for nvidia-container-cli
+	// ldconfig needs to be run in the container to setup the correct symlinks to the library files nvidia-container-cli
+	// maps into the container
+	args := []string{nvidiaToolPath, debugOption, "--load-kmods", "--no-pivot", "configure", "--ldconfig=@/sbin/ldconfig"}
+	if capabilities, ok := spec.Annotations[annotationContainerGPUCapabilities]; ok {
+		caps := strings.Split(capabilities, ",")
+		for _, c := range caps {
+			args = append(args, fmt.Sprintf("--%s", c))
+		}
+	}
+
+	for _, d := range spec.Windows.Devices {
+		switch d.IDType {
+		case "gpu":
+			busLocation, err := pci.FindDeviceBusLocationFromVMBusGUID(ctx, d.ID)
+			if err != nil {
+				return errors.Wrapf(err, "failed to find nvidia gpu bus location")
+			}
+			args = append(args, fmt.Sprintf("--device=%s", busLocation))
+		}
+	}
+
+	args = append(args, "--no-cgroups", "--pid=%v", spec.Root.Path)
+
+	if spec.Hooks == nil {
+		spec.Hooks = &oci.Hooks{}
+	}
+
+	nvidiaHook := oci.Hook{
+		Path: nvidiaToolPath,
+		Args: args,
+		Env:  updateEnvWithNvidiaVariables(),
+	}
+
+	spec.Hooks.Prestart = append(spec.Hooks.Prestart, nvidiaHook)
+	return nil
+}
+
+// updateEnvWithNvidiaVariables creates an env with the nvidia gpu vhd in PATH and insecure mode set
+func updateEnvWithNvidiaVariables() []string {
+	pathPrefix := "PATH="
+	nvidiaBin := fmt.Sprintf("%s/bin", lcowNvidiaMountPath)
+	env := os.Environ()
+	for i, v := range env {
+		if strings.HasPrefix(v, pathPrefix) {
+			newPath := fmt.Sprintf("%s:%s", v, nvidiaBin)
+			env[i] = newPath
+		}
+	}
+	// NVC_INSECURE_MODE allows us to run nvidia-container-cli without seccomp
+	// we don't currently use seccomp in the uvm, so avoid using it here for now as well
+	env = append(env, "NVC_INSECURE_MODE=1")
+	return env
+}
--- a/internal/runtime/hcsv2/uvm.go
+++ b/internal/runtime/hcsv2/uvm.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/Microsoft/opengcs/internal/storage"
 	"github.com/Microsoft/opengcs/internal/storage/overlay"
+	"github.com/Microsoft/opengcs/internal/storage/pci"
 	"github.com/Microsoft/opengcs/internal/storage/plan9"
 	"github.com/Microsoft/opengcs/internal/storage/pmem"
 	"github.com/Microsoft/opengcs/internal/storage/scsi"
@@ -204,6 +205,8 @@ func (h *Host) ModifyHostSettings(ctx context.Context, settings *prot.ModifySett
 		return modifyCombinedLayers(ctx, settings.RequestType, settings.Settings.(*prot.CombinedLayersV2))
 	case prot.MrtNetwork:
 		return modifyNetwork(ctx, settings.RequestType, settings.Settings.(*prot.NetworkAdapterV2))
+	case prot.MrtVPCIDevice:
+		return modifyMappedVPCIDevice(ctx, settings.RequestType, settings.Settings.(*prot.MappedVPCIDeviceV2))
 	default:
 		return errors.Errorf("the ResourceType \"%s\" is not supported", settings.ResourceType)
 	}
@@ -357,6 +360,15 @@ func modifyMappedVPMemDevice(ctx context.Context, rt prot.ModifyRequestType, vpd
 		return pmem.Mount(ctx, vpd.DeviceNumber, vpd.MountPath)
 	case prot.MreqtRemove:
 		return storage.UnmountPath(ctx, vpd.MountPath, true)
+	default:
+		return newInvalidRequestTypeError(rt)
+	}
+}
+
+func modifyMappedVPCIDevice(ctx context.Context, rt prot.ModifyRequestType, vpciDev *prot.MappedVPCIDeviceV2) error {
+	switch rt {
+	case prot.MreqtAdd:
+		return pci.WaitForPCIDeviceFromVMBusGUID(ctx, vpciDev.VMBusGUID)
 	default:
 		return newInvalidRequestTypeError(rt)
 	}
--- a/internal/runtime/hcsv2/workload_container.go
+++ b/internal/runtime/hcsv2/workload_container.go
@@ -43,6 +43,15 @@ func updateSandboxMounts(sbid string, spec *oci.Spec) error {
 	return nil
 }
 
+func specHasGPUDevice(spec *oci.Spec) bool {
+	for _, d := range spec.Windows.Devices {
+		if d.IDType == "gpu" {
+			return true
+		}
+	}
+	return false
+}
+
 func setupWorkloadContainerSpec(ctx context.Context, sbid, id string, spec *oci.Spec) (err error) {
 	ctx, span := trace.StartSpan(ctx, "hcsv2::setupWorkloadContainerSpec")
 	defer span.End()
@@ -162,6 +171,13 @@ func setupWorkloadContainerSpec(ctx context.Context, sbid, id string, spec *oci.
 
 	// Force the parent cgroup into our /containers root
 	spec.Linux.CgroupsPath = "/containers/" + id
+
+	if spec.Windows != nil && specHasGPUDevice(spec) {
+		// we only support Nvidia gpus right now
+		if err := addNvidiaDevicePreHook(ctx, spec); err != nil {
+			return err
+		}
+	}
 
 	// Clear the windows section as we dont want to forward to runc
 	spec.Windows = nil
--- a/internal/storage/pci/pci.go
+++ b/internal/storage/pci/pci.go
@@ -1,0 +1,64 @@
+// +build linux
+
+package pci
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/Microsoft/opengcs/internal/storage"
+	"github.com/Microsoft/opengcs/internal/storage/vmbus"
+)
+
+var storageWaitForFileMatchingPattern = storage.WaitForFileMatchingPattern
+var vmbusWaitForDevicePath = vmbus.WaitForDevicePath
+
+// WaitForPCIDeviceFromVMBusGUID waits for bus location path of the device to be present
+func WaitForPCIDeviceFromVMBusGUID(ctx context.Context, vmBusGUID string) error {
+	_, err := FindDeviceBusLocationFromVMBusGUID(ctx, vmBusGUID)
+	return err
+}
+
+// FindDeviceBusLocationFromVMBusGUID finds device bus location by
+// reading /sys/bus/vmbus/devices/<vmBusGUID>/... for pci specific directories
+func FindDeviceBusLocationFromVMBusGUID(ctx context.Context, vmBusGUID string) (string, error) {
+	pciDir, err := findVMBusPCIDir(ctx, vmBusGUID)
+	if err != nil {
+		return "", err
+	}
+
+	pciDeviceLocation, err := findVMBusPCIDevice(ctx, pciDir)
+	if err != nil {
+		return "", err
+	}
+	return pciDeviceLocation, nil
+}
+
+// findVMBusPCIDir waits for the pci bus directory matching pattern
+// /sys/bus/vmbus/devices/<vmBusGUID>/pci* to exist and returns
+// the full resulting path or an error
+func findVMBusPCIDir(ctx context.Context, vmBusGUID string) (string, error) {
+	vmBusPCIPathPattern := filepath.Join(vmBusGUID, "pci*")
+	return vmbusWaitForDevicePath(ctx, vmBusPCIPathPattern)
+}
+
+// findVMBusPCIDevice waits for the pci bus location directory under the path
+// returned from findVMBusPCIDir to exist and returns the pci bus location or an error
+func findVMBusPCIDevice(ctx context.Context, pciDirFullPath string) (string, error) {
+	// trim /sys/bus/vmbus/devices/<vmBusGUID>/pciXXXX:XX to XXXX:XX
+	_, pciDirName := filepath.Split(pciDirFullPath)
+	busPrefix := strings.TrimPrefix(pciDirName, "pci")
+
+	// under /sys/bus/vmbus/devices/<vmBusGUID>/pciXXXX:XX/ look for directory matching XXXX:XX* pattern
+	busPathPattern := filepath.Join(pciDirFullPath, fmt.Sprintf("%s*", busPrefix))
+	busFileFullPath, err := storageWaitForFileMatchingPattern(ctx, busPathPattern)
+	if err != nil {
+		return "", err
+	}
+
+	// return the resulting XXXX:XX:YY.Y pci bus location
+	_, busFile := filepath.Split(busFileFullPath)
+	return busFile, nil
+}
--- a/internal/storage/utilities.go
+++ b/internal/storage/utilities.go
@@ -1,0 +1,38 @@
+// +build linux
+
+package storage
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// export this variable so it can be mocked to aid in testing for consuming packages
+var filepathglob = filepath.Glob
+
+// WaitForFileMatchingPattern waits for a single file that matches the given path pattern and returns the full path
+// to the resulting file
+func WaitForFileMatchingPattern(ctx context.Context, pattern string) (string, error) {
+	for {
+		files, err := filepathglob(pattern)
+		if err != nil {
+			return "", err
+		}
+		if len(files) == 0 {
+			select {
+			case <-ctx.Done():
+				return "", errors.Wrapf(ctx.Err(), "timed out waiting for file matching pattern %s to exist", pattern)
+			default:
+				time.Sleep(time.Millisecond * 10)
+				continue
+			}
+		} else if len(files) > 1 {
+			return "", fmt.Errorf("more than one file could exist for pattern \"%s\"", pattern)
+		}
+		return files[0], nil
+	}
+}
--- a/internal/storage/vmbus/vmbus.go
+++ b/internal/storage/vmbus/vmbus.go
@@ -1,0 +1,18 @@
+// +build linux
+
+package vmbus
+
+import (
+	"context"
+	"path/filepath"
+
+	"github.com/Microsoft/opengcs/internal/storage"
+)
+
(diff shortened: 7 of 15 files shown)


internal/runtime/hcsv2/workload_container.go
```diff
@@ -163,8 +167,67 @@ func setupWorkloadContainerSpec(ctx context.Context, sbid, id string, spec *oci.
 	// Force the parent cgroup into our /containers root
 	spec.Linux.CgroupsPath = "/containers/" + id
 
+	if spec.Windows != nil && len(spec.Windows.Devices) > 0 {
```
jterry75: Just do this check in `addNvidiaDevicePreHook`

internal/runtime/hcsv2/workload_container.go
```diff
@@ -163,8 +167,67 @@ func setupWorkloadContainerSpec(ctx context.Context, sbid, id string, spec *oci.
...
+		caps := strings.Split(capabilities, ",")
+		for _, c := range caps {
+			args = append(args, fmt.Sprintf("--%s", c))
+		}
+	}
+
+	for _, d := range spec.Windows.Devices {
+		switch d.IDType {
+		case "gpu":
+			args = append(args, fmt.Sprintf("--device=%s", d.ID))
```
jterry75: Isn't this the host vpci path?

internal/runtime/hcsv2/workload_container.go
```diff
@@ -14,6 +16,8 @@ import (
 	"go.opencensus.io/trace"
 )
 
+const lcowNvidiaMountPath = "/nvidia"
```
jterry75: Should this be under `/run` like all other gcs state?

service/gcsutils/gcstools/nvidiaPrestartHook.go
```diff
@@ -0,0 +1,71 @@
+package main
```
jterry75: Interesting. I had assumed this tool would be part of the vhd that gets mounted in. Do we want it as a UVM builtin?

internal/runtime/hcsv2/workload_container.go
katiewasnothere: The reasoning behind this is [here](https://github.com/microsoft/hcsshim/pull/765#discussion_r373601465) I believe

internal/runtime/hcsv2/workload_container.go
katiewasnothere: I'm fine with that change

service/gcsutils/gcstools/nvidiaPrestartHook.go
katiewasnothere: I don't see a big issue with putting it in the vhd. The only concern would be that if we wanted to say update the lcowNvidiaMountPath we'd have to have LSG update the vhd image.

internal/runtime/hcsv2/workload_container.go
katiewasnothere: I feel like that change would make the code a bit less understandable here, since we aren't always wanting to add an nvidia prestart hook. Do you feel strongly about this?

katiewasnothere:
@jterry75 @kevpar @jstarks please take another look here as well

internal/runtime/hcsv2/uvm.go
```diff
@@ -196,6 +197,8 @@ func (h *Host) ModifyHostSettings(ctx context.Context, settings *prot.ModifySett
 		return modifyCombinedLayers(ctx, settings.RequestType, settings.Settings.(*prot.CombinedLayersV2))
 	case prot.MrtNetwork:
 		return modifyNetwork(ctx, settings.RequestType, settings.Settings.(*prot.NetworkAdapterV2))
+	case prot.MrtVPciDevice:
+		return modifyMappedVPciDevice(ctx, settings.RequestType, settings.Settings.(*prot.MappedVPciDeviceV2))
```
jterry75: Love it!

internal/runtime/hcsv2/workload_container.go
jterry75: I'm saying that len(spec.Windows.Devices) is not a good enough check. So its an odd pre-check. I get the spec.Windows != nil part thats obvious. But really there needs to be spec.Windows.Device[i] where IDType == "gpu" so seems weird to have this check here. Alternatively do the loop iteration here and then call the perHook when you find a matching device so that its more obvious.

internal/storage/pci/pci.go
```diff
@@ -0,0 +1,84 @@
...
+	pciDeviceLocation, err := findVmbusPciDevice(ctx, pciDir)
+	if err != nil {
+		return "", err
+	}
+	return pciDeviceLocation, nil
+}
+
+// findVmbusPciDevice waits for the pci bus location directory under the path
+// returned from findVmbusPciDirs to exist and returns the full pci bus location or an error
+func findVmbusPciDevice(ctx context.Context, pciDirFullPath string) (string, error) {
```
jterry75: I really like this package! I know it seems a little crazy but you could avoid this double indirection if we make a vmbus package and a pci package here. And in fact that vmbus package would super helpful for other things as well like nic's. 

So you could make vmbus.WaitForDevice(ctx, id string)
Which waits for /sys/bus/vmbus/devices/<id> or timeout. 

Then the pci package is WaitForDevice(ctx, id string) which waits for vmbus.WaitForDevice and then does the /pci* work.

katiewasnothere: I'm good with that change! Could you explain a bit why the vmbus change would be helpful for nic's?

jterry75: Yes. Because this code: https://github.com/microsoft/opengcs/blob/master/internal/network/network.go#L114 is actually just vmbus.WaitForDevice followed by "wait for the net" dir. So we can only have that logic in one place

internal/runtime/hcsv2/nvidia_utils.go
```diff
@@ -0,0 +1,77 @@
...
+
+// addNvidiaDevicePreHook builds the arguments for nvidia-container-cli and creates the prestart hook
+func addNvidiaDevicePreHook(ctx context.Context, spec *oci.Spec) error {
+	nvidiaToolBinary := "nvidiaPrestartHook"
+	nvidiaToolPath, err := exec.LookPath(nvidiaToolBinary)
+	if err != nil {
+		return errors.Wrapf(err, "failed to find %s for container GPU support", nvidiaToolBinary)
+	}
+
+	var args []string
```
kevpar: FYI you can also write this as:

```
args := []string{nvidiaToolPath, "--debug=/nvidia-container.log", etc}
```

internal/runtime/hcsv2/nvidia_utils.go
```diff
@@ -0,0 +1,77 @@
...
+	args = append(args, nvidiaToolPath, "--debug=/nvidia-container.log", "--load-kmods", "--no-pivot", "configure", "--ldconfig=/sbin/ldconfig")
+	if capabilities, ok := spec.Annotations["io.microsoft.container.gpu.capabilities"]; ok {
```
kevpar: We should make a variable for this annotation name.

internal/runtime/hcsv2/nvidia_utils.go
```diff
@@ -0,0 +1,77 @@
...
+			busLocation, err := pci.FindDeviceBusLocationFromVMBusGUID(ctx, d.ID)
+			if err != nil {
+				return errors.Wrapf(err, "failed to find nvidia gpu bus location")
+			}
+			args = append(args, fmt.Sprintf("--device=%s", busLocation))
+		}
+	}
+
+	rootfsDir := spec.Root.Path
+	args = append(args, "--no-cgroups", "--pid=%v", fmt.Sprintf("%s", rootfsDir))
```
kevpar: Do we need to do this double-printf formatting here?

internal/runtime/hcsv2/nvidia_utils.go
```diff
@@ -0,0 +1,77 @@
...
+		spec.Hooks = &oci.Hooks{}
+	}
+
+	nvidiaHook := oci.Hook{
+		Path: nvidiaToolPath,
+		Args: args,
+		Env:  updateEnvWithNvidiaPath(),
+	}
+
+	spec.Hooks.Prestart = append(spec.Hooks.Prestart, nvidiaHook)
```
kevpar: I seem to remember something about `Prestart` being deprecated. Is that the case? Should we add a comment here about it?

internal/runtime/hcsv2/nvidia_utils.go
```diff
@@ -0,0 +1,77 @@
...
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/Microsoft/opengcs/internal/storage/pci"
+	oci "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+)
+
+const lcowNvidiaMountPath = "/run/nvidia"
```
kevpar: I would add a comment here that this needs to match the path that is mapped to by the shim.

internal/runtime/hcsv2/nvidia_utils.go
```diff
@@ -0,0 +1,77 @@
...
+	pathPrefix := "PATH="
+	nvidiaBin := fmt.Sprintf("%s/bin", lcowNvidiaMountPath)
+	env := os.Environ()
+	for i, v := range env {
+		if strings.HasPrefix(v, pathPrefix) {
+			newPath := fmt.Sprintf("%s:%s", v, nvidiaBin)
+			env[i] = newPath
+		}
+	}
+	env = append(env, "NVC_INSECURE_MODE=1")
```
kevpar: What is this?

internal/runtime/hcsv2/uvm.go
```diff
@@ -354,6 +357,15 @@ func modifyMappedVPMemDevice(ctx context.Context, rt prot.ModifyRequestType, vpd
 	}
 }
 
+func modifyMappedVPCIDevice(ctx context.Context, rt prot.ModifyRequestType, vpciDev *prot.MappedVPCIDeviceV2) error {
+	switch rt {
```
kevpar: Do we need to handle the remove case here?

internal/storage/pci/pci.go
```diff
@@ -0,0 +1,84 @@
...
+	pciDeviceLocation, err := findVMBusPCIDevice(ctx, pciDir)
+	if err != nil {
+		return "", err
+	}
+	return pciDeviceLocation, nil
+}
+
+// findVMBusPCIDevice waits for the pci bus location directory under the path
+// returned from findVMBusPCIDirs to exist and returns the full pci bus location or an error
+func findVMBusPCIDevice(ctx context.Context, pciDirFullPath string) (string, error) {
```
kevpar: I think a comment in here explaining the logic of the paths you are examining would be good. Especially if it includes an example of the paths so we can follow along.

internal/runtime/hcsv2/nvidia_utils.go
katiewasnothere: Its use is going to be deprecated, but the new [hooks described in the runtime spec](https://github.com/opencontainers/runtime-spec/blob/39c287c415bf86fb5b7506528d471db5405f8ca8/config.md#posix-platform-hooks) have not yet been implemented in runc. 

Looks like there is active work and [discussion](https://github.com/opencontainers/runc/pull/1811#issuecomment-582682968) to implement the new hooks, but it's unclear when that will be done. 

I will add a comment to indicate we should update to the new hooks asap.

internal/runtime/hcsv2/nvidia_utils.go
katiewasnothere: Will do. Do you know of a better way to keep these two things synced?

kevpar: We could put this value in a package that is used in both places, but that seems like overkill to me.

internal/runtime/hcsv2/nvidia_utils.go
katiewasnothere: nvidia-container-cli typically uses seccomp to limit syscalls. In order to prevent using seccomp, the tool expects this environment variable to [be set](https://github.com/NVIDIA/libnvidia-container/blob/6ed0f129eb63142ce60a57e7657638b552327a85/src/nvc_ldcache.c#L41) as well as the package to have been [compiled without seccomp support](https://github.com/NVIDIA/libnvidia-container/blob/6ed0f129eb63142ce60a57e7657638b552327a85/Makefile#L12)

kevpar: Cool. Let's add a comment here saying that, and rename the function this is in so it's clear it's doing more than just setting PATH.

internal/runtime/hcsv2/uvm.go
katiewasnothere: In the gpu case goes, I don't think so. AFAIK nvidia-container-cli doesn't support removing a gpu from a container, so the only removal that needs to be done is of the vpci device on the uvm.

internal/runtime/hcsv2/nvidia_utils.go
katiewasnothere: Yeah I agree. We'll keep it this way then for now.

internal/storage/pci/pci.go
katiewasnothere: I would still need to wait twice in the pci package, not including the wait that would be in the vmbus package, since I have to wait for a pci* directory to exist under /sys/bus/vmbus/devices/<vmbus guid> and then separately wait for a <pci bus location> directory under that. 

For example, a gpu with a guid of 1111-2222-3333-4444 and pci bus location 1234:00.00.0, we need to wait for:
/sys/bus/vmbus/devices/1111-2222-3333-4444/pci*
and /sys/bus/vmbus/devices/1111-2222-3333-4444/pci1234:00/1234:00*

internal/storage/vmbus/vmbus.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+	"path/filepath"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// WaitForVMBusGUIDPath waits for the vmbus device to exist at /sys/bus/vmbus/devices/<vmBusGUID>
+func WaitForVMBusGUIDPath(ctx context.Context, vmBusGUID string) (string, error) {
+	vmBusGUIDPath := fmt.Sprintf("/sys/bus/vmbus/devices/%s*", vmBusGUID)
```
jterry75: At this level of the package it will always be exactly `%s` right? And then its not a `filepath.Glob` its a `filepath.Stat` right?

katiewasnothere: That's true, though keeping filepath.Glob allows for more code reuse with WaitForFileMatchingPattern. I can make a separate func for stat'ing the vmbus path if you'd prefer.

jterry75: But I am saying your func here is asking for the `vmBusGUID` which is not a `pattern` at all. I am ok with making it more generic if you like so that we dont need to wait for the various 2nd level dirs in say networking and PCI but then you probably want to make the vmbus.WaitForDevice(ctx, vmBusGUIDPattern string)` so someone can pass `<GUID>/net` and this code can wait for the entire sub tree. Up to you I guess how you would like to unify here.

katiewasnothere: I made this change. It doesn't stop us from having to wait twice in the network scenario (since we need to wait for /sys/bus..../<guid>/net to exist and additionally for a target directory under that to exist), but it does cut down waiting for the pci case to two :)

kevpar:
Have you thought much about what diagnostics we will need to get for the GPU setup? From some brief code searching, it appears that any output we log from the prestart hook will be re-logged from runc, so hopefully that will end up in our existing runc error logging mechanism (but you should test this). Are there any other logs generated from the nvidia tool we should be grabbing on failure?

katiewasnothere:
@kevpar the nvidia tool logs output to a specified file, which I've set to be /nvidia-container.log [here](https://github.com/microsoft/opengcs/blob/4733c0226ee1530dadeb359a7771d460d02defe4/internal/runtime/hcsv2/nvidia_utils.go#L35). Their logs don't look like they can be directly consumed by logrus, so there will need to be work around that, though we could also dump that file with logrus. 

As for other diagnostics, I'm not sure what we would want or need. Do you have other suggestions?

internal/storage/pci/pci.go
katiewasnothere: Do the new comments help?

internal/runtime/hcsv2/workload_container.go
katiewasnothere: @jterry75 is the new check okay?

internal/runtime/hcsv2/nvidia_utils.go
katiewasnothere: Done!

internal/storage/pci/pci.go
katiewasnothere: do the new comments make it clearer?

katiewasnothere:
> I think we should grab that file and log its output (maybe in chunks if it's big) in the case of a failure. In production we won't be able to connect to the UVM in a lot of cases to diagnose, so need to make sure we grab anything that we would need.

I think that the file should be pretty small since it looks like roughly the same number of log statements are generated for any given configure command.

internal/storage/vmbus/vmbus.go
katiewasnothere: @jterry75 is this change good?

service/gcsutils/gcstools/nvidiaPrestartHook.go
```diff
@@ -0,0 +1,91 @@
...
+	numBytesInContents := len(contents)
+
+	// don't try to log over 8KB to avoid issues with etw
+	stopBytes := 8000
+	if numBytesInContents < stopBytes {
+		stopBytes = numBytesInContents
+	}
+	output := string(contents[:stopBytes])
+	logrus.WithField("output", output).Info("nvidia-container-cli debug output")
+}
```
katiewasnothere: @kevpar please take a look

internal/storage/vmbus/vmbus.go
```diff
@@ -11,8 +11,10 @@ import (
 	"github.com/pkg/errors"
 )
 
-// WaitForVMBusDevicePath waits for the vmbus device to exist at /sys/bus/vmbus/devices/<vmbusGUIDPattern>...
-func WaitForVMBusDevicePath(ctx context.Context, vmbusGUIDPattern string) (string, error) {
+var WaitForVMBusDevicePath = waitForVMBusDevicePath
```
jterry75: I typically try to only 'fake' the part that does a call I cant control. So basically in this case its the `filepath.Glob` method. So I would make a 

`var filepathGlob = filepath.Glob`
and then in `WaitForVMBusDevicePath` call `filepathGlob(pattern)`. In the unit tests you can now control the response of `files` or `err` and you can handle all other cases via cancellation etc. 

Doing it the way you did there actually leaks the ability to change the `WaitForVMBusDevicePath` method for anyone in the binary which seems odd.

katiewasnothere: That makes sense. Just updated with these changes in the pci tests.

internal/storage/vmbus/vmbus.go
```diff
@@ -0,0 +1,43 @@
...
+func WaitForVMBusDevicePath(ctx context.Context, vmbusGUIDPattern string) (string, error) {
+	vmBusPath := filepath.Join("/sys/bus/vmbus/devices", vmbusGUIDPattern)
+	return WaitForFileMatchingPattern(ctx, vmBusPath)
+}
+
+// WaitForFileMatchingPattern waits for a single file that matches the given path pattern and returns the full path
+// to the resulting file
+func WaitForFileMatchingPattern(ctx context.Context, pattern string) (string, error) {
+	for {
+		files, err := Filepathglob(pattern)
```
kevpar: Can we just write `filepath.Glob` here?

internal/storage/vmbus/vmbus.go
```diff
@@ -0,0 +1,43 @@
...
+
+// WaitForVMBusDevicePath waits for the vmbus device to exist at /sys/bus/vmbus/devices/<vmbusGUIDPattern>...
+func WaitForVMBusDevicePath(ctx context.Context, vmbusGUIDPattern string) (string, error) {
+	vmBusPath := filepath.Join("/sys/bus/vmbus/devices", vmbusGUIDPattern)
+	return WaitForFileMatchingPattern(ctx, vmBusPath)
+}
+
+// WaitForFileMatchingPattern waits for a single file that matches the given path pattern and returns the full path
+// to the resulting file
+func WaitForFileMatchingPattern(ctx context.Context, pattern string) (string, error) {
```
kevpar: I love that we have this helper function now, but think it should be broken out to a utility or filesystem related package somewhere, rather than be exported from `vmbus`.

internal/storage/vmbus/vmbus.go
katiewasnothere: Justin's comment above explains why: https://github.com/microsoft/opengcs/pull/369#discussion_r379255089

internal/runtime/hcsv2/nvidia_utils.go
```diff
@@ -0,0 +1,87 @@
...
+// addNvidiaDevicePreHook builds the arguments for nvidia-container-cli and creates the prestart hook
+func addNvidiaDevicePreHook(ctx context.Context, spec *oci.Spec) error {
+	nvidiaToolBinary := "nvidiaPrestartHook"
+	nvidiaToolPath, err := exec.LookPath(nvidiaToolBinary)
+	if err != nil {
+		return errors.Wrapf(err, "failed to find %s for container GPU support", nvidiaToolBinary)
+	}
+
+	debugOption := fmt.Sprintf("--debug=%s", nvidiaDebugFilePath)
+	args := []string{nvidiaToolPath, debugOption, "--load-kmods", "--no-pivot", "configure", "--ldconfig=@/sbin/ldconfig"}
```
kevpar: I remember the situation with ldconfig and whether we use the version on the host or in the container to be somewhat complicated. Can you add a comment explaining the reasoning behind what you're doing with that?

internal/storage/vmbus/vmbus.go
katiewasnothere: I agree. I can make that change and update the vmbus package tests to be a small test similar to the pci package's tests.

internal/runtime/hcsv2/nvidia_utils.go
katiewasnothere: Both the host and container ldconfig don't work as intended for the nvidia-container-cli tool here. So really until I can investigate that issue more, this is just arbitrary. 

Since this is still an open issue, the workaround is just to run ldconfig as part of the container config's command.

internal/runtime/hcsv2/nvidia_utils.go
```diff
@@ -0,0 +1,87 @@
...
+	pathPrefix := "PATH="
+	nvidiaBin := fmt.Sprintf("%s/bin", lcowNvidiaMountPath)
+	env := os.Environ()
+	for i, v := range env {
+		if strings.HasPrefix(v, pathPrefix) {
+			newPath := fmt.Sprintf("%s:%s", v, nvidiaBin)
+			env[i] = newPath
+		}
+	}
+	// NVC_INSECURE_MODE allows us to run nvidia-container-cli without seccomp
```
kevpar: Let's expand this comment to explain why we can't/don't want to run with seccomp

internal/storage/vmbus/vmbus.go
kevpar: Oh, I see it's exported and we're using it for some sort of test dependency injection.

internal/storage/vmbus/vmbus.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+var Filepathglob = filepath.Glob
```
kevpar: Exporting this so we can replace it for testing seems a bit gross to me. If we can find a cleaner way to do it, that would be nice. If not, let's at least clearly document here why this is exposed this way.

service/gcsutils/gcstools/nvidiaPrestartHook.go
```diff
@@ -0,0 +1,91 @@
...
+
+func logDebugFile() {
+	contents, err := ioutil.ReadFile(nvidiaDebugFilePath)
+	if err != nil {
+		logrus.Errorf("failed to read nvidia-container-cli debug file: %s", err)
+		return
+	}
+	numBytesInContents := len(contents)
+
+	// don't try to log over 8KB to avoid issues with etw
```
kevpar: I would add something like "We generally forward logging to Windows where it is logged out as ETW..."

service/gcsutils/gcstools/nvidiaPrestartHook.go
```diff
@@ -0,0 +1,91 @@
...
+	stopBytes := 8000
+	if numBytesInContents < stopBytes {
+		stopBytes = numBytesInContents
+	}
+	output := string(contents[:stopBytes])
```
kevpar: We should loop and log out the file in 8KB chunks, rather than just stopping when we hit the limit.

internal/runtime/hcsv2/nvidia_utils.go
kevpar: If this isn't resolved prior to checking in, please make sure you add a comment explaining this.

internal/storage/vmbus/vmbus.go
kevpar: Found the comment with Justin where he recommended doing it this way. :) I think that's fine, testing is always grosser than I want it to be. Please just add a comment here explaining.

internal/storage/vmbus/vmbus.go
katiewasnothere: Done!

internal/storage/vmbus/vmbus.go
katiewasnothere: Done!

internal/storage/utilities.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// stub this so we can make unit tests for depending packages
```
kevpar: I would be a bit more explicit here, something like "we export this as a variable here so that it can be mocked from other packages to help with testing"

service/gcsutils/gcstools/nvidiaPrestartHook.go
katiewasnothere: done

internal/network/network.go
```diff
@@ -111,7 +112,9 @@ func InstanceIDToName(ctx context.Context, id string) (_ string, err error) {
 	id = strings.ToLower(id)
 	span.AddAttributes(trace.StringAttribute("adapterInstanceID", id))
 
-	devicePath := filepath.Join("/sys", "bus", "vmbus", "devices", id, "net")
+	vmBusSubPath := filepath.Join(id, "net")
+	devicePath, err := vmbus.WaitForVMBusDevicePath(ctx, vmBusSubPath)
```
jterry75: NIT: since this is the `vmbus` package it should just be `WaitForDevicePath`

internal/runtime/hcsv2/nvidia_utils.go
```diff
@@ -0,0 +1,92 @@
...
+		case "gpu":
+			busLocation, err := pci.FindDeviceBusLocationFromVMBusGUID(ctx, d.ID)
+			if err != nil {
+				return errors.Wrapf(err, "failed to find nvidia gpu bus location")
+			}
+			args = append(args, fmt.Sprintf("--device=%s", busLocation))
+		}
+	}
+
+	rootfsDir := spec.Root.Path
```
jterry75: There are no modifications to `rootfsDir` why not just pass `spec.Root.Path` below?

internal/storage/utilities.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+	"fmt"
+	"path/filepath"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// export this variable so it can be mocked to aid in testing for consuming packages
+var Filepathglob = filepath.Glob
```
jterry75: I'm confused why you are exporting these. The test package as access to this unexported so it doesnt need to be exported to mock.

internal/runtime/hcsv2/nvidia_utils.go
katiewasnothere: I feel like I remember Kevin leaving this exact same comment. Maybe I missed something. I'll update this :)

internal/storage/utilities.go
katiewasnothere: Because I also mock it out for my pci package test :)

PR #371, opened 2020-03-03
Terminate all other container processes when the init exits

kevpar:
Previously, if the container init spawned a child process which
inherited its stdio, then exited, gcs would continue waiting until all
child processes terminated. This only occurs when the container does not
have its own pid namespace, as otherwise the kernel terminates all
processes in the namespace automatically. To preserve this semantic even
when the namespace is shared, we now explicitly kill all other container
processes when its init exits.

Signed-off-by: Kevin Parsons <kevpar@microsoft.com>

### DIFF
--- a/service/gcs/runtime/runc/runc.go
+++ b/service/gcs/runtime/runc/runc.go
@@ -20,6 +20,7 @@ import (
 	containerdsys "github.com/containerd/containerd/sys"
 	oci "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 const (
@@ -39,6 +40,9 @@ type container struct {
 	r    *runcRuntime
 	id   string
 	init *process
+	// ownsPidNamespace indicates whether the container's init process is also
+	// the init process for its pid namespace.
+	ownsPidNamespace bool
 }
 
 func (c *container) ID() string {
@@ -57,6 +61,8 @@ func (c *container) PipeRelay() *stdio.PipeRelay {
 	return c.init.pipeRelay
 }
 
+// process represents a process running in a container. It can either be a
+// container's init process, or an exec process in a container.
 type process struct {
 	c         *container
 	pid       int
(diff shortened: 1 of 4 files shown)


service/gcs/runtime/runc/runc.go
```diff
@@ -427,6 +433,37 @@ func (r *runcRuntime) waitOnProcess(pid int) (int, error) {
...
+	l := logrus.WithField("cid", p.c.id)
+	l.WithField("pid", p.pid).Debug("process wait completed")
+
+	// If the init process for the container has exited, kill everything else in
+	// the container. We don't do this if the container owns its own pid
+	// namespace, because in that case the container init will be the pid
+	// namespace init, so the kernel will kill everything automatically when it
+	// exits. Excluding this condition allows a use case where two containers
+	// can share cgroups (which is how runc determines what to kill), but still
+	// have their processes terminated properly, by not sharing a pid namespace.
```
katiewasnothere: Nit: I feel like this last sentence is kind of confusing the way its written.

service/gcs/runtime/runc/runc.go
```diff
@@ -473,27 +514,35 @@ func (r *runcRuntime) runCreateCommand(id string, bundlePath string, stdioSet *s
...
-				// Intentionally ignore the error.
+	// Determine if the container owns its own pid namespace or not. Per the OCI
+	// spec:
+	// - If the spec has no entry for the pid namespace, the container inherits
+	//   the runtime namespace (container does not own).
+	// - If the spec has a pid namespace entry, but the path is empty, a new
+	//   namespace will be created and used for the container (container owns).
+	// - If there is a pid namespace entry with a path, the container uses the
+	//   namespace at that path (container does not own).
+	if spec.Linux == nil {
```
katiewasnothere: did you mean != nil?

service/gcs/runtime/runc/runc.go
kevpar: Thoughts on how it could be improved? What did you find confusing about it?

service/gcs/runtime/runc/runc.go
kevpar: Yes :)

service/gcs/runtime/runc/runc.go
katiewasnothere: I assume what you're trying to say is that without this condition it's still possible that processes are terminated properly in the case that two containers share a cgroup iff they do not also share a pid namespace. Is that correct?

jstarks: How about:

That way, if two containers share cgroups but not a pid namespace, then when one container terminates, it will not cause the other container to terminate.

And maybe mention that runc is using the cgroup as the determination up above.

kevpar: I've rephrased the comment. @katiewasnothere is it clearer now?

katiewasnothere: I still think it's a bit confusing but it's not super crucial so we can merge if you think this is best. But I think something like "This prevents the issue where if two containers share cgroups but not a pid namespace (they each own their pid namespace), one container's termination would call runc kill and kill both container's processes" makes it more clear what the original issue was. Are you okay with that?

kevpar: How about this?

```
	// If the init process for the container has exited, kill everything else in
	// the container. Runc uses the devices cgroup of the container ot determine
	// what other processes to kill.
	//
	// We don't issue the kill if the container owns its own pid namespace,
	// because in that case the container kernel will kill everything in the pid
	// namespace automatically (as the container init will be the pid namespace
	// init). This prevents a potential issue where two containers share cgroups
	// but have their own pid namespaces. If we didn't handle this case, runc
	// would kill the processes in both containers.
```

kevpar: ^ Slight edit

katiewasnothere: Works for me :)

katiewasnothere:
@kevpar I don't think you pushed the most recent comment change

kevpar:
Test for this: https://github.com/microsoft/hcsshim/pull/778

PR #372, opened 2020-03-23
Improve low memory detection

dcantah:
* Remove 50 MB default hard limit on the gcs cgroup.
* Register an eventfd to get notified when the cgroup the gcs is in
goes over 50 MB memory threshold.
* Register for oom notification on the containers cgroup.


service/gcs/main.go
```diff
@@ -18,8 +20,24 @@ import (
...
 
+func readMemoryEvents(efd uintptr, cgName string, limit int64) {
+	// Buffer must be >= 8 bytes for eventfd reads
+	// http://man7.org/linux/man-pages/man2/eventfd.2.html
+	buf := make([]byte, 8)
+	for {
+		_, err := cgrouputils.ReadEventFD(efd, buf)
+		if err != nil {
+			logrus.WithError(err).Warn("opengcs::readEvents - failed to read from eventfd")
+			unix.Close(int(efd))
```
jstarks: Should this just be a defer at the top of the function?

It's a little weird that this function takes ownership of the fd; that should be made clear in a comment.

service/gcs/main.go
```diff
@@ -18,8 +20,24 @@ import (
...
+			return
+		}
+		logrus.Warnf("opengcs::readMemoryEvents - memory usage for cgroup '%s' exceeded %d byte threshold", cgName, limit)
```
jstarks: This seems sufficient for the gcs case, but for the container OOM case is there a way for us to get more detailed info?

service/gcs/main.go
```diff
@@ -150,6 +164,18 @@ func main() {
...
+	if err != nil {
+		logrus.WithError(err).Fatal("opengcs::main - failed to register memory threshold for gcs cgroup")
+	}
+
+	containerEfd, err := containersControl.OOMEventFD()
+	if err != nil {
+		logrus.WithError(err).Fatal("opengcs::main - failed to retrieve the container cgroups oom eventfd")
+	}
+
+	go readMemoryEvents(gcsEfd, "gcs", gcsLimit)
```
jstarks: Can we add some tests for these two scenarios? The tests might have to be in the hcsshim repo. This could be a followup change.

service/gcs/main.go
jstarks: Even for GCS it would be good to get the memory use, perhaps the high water mark after a second or something?

dcantah: Sounds good

service/gcs/main.go
dcantah: Yes, I didn't know of how to test this in this repo directly but that can be a good test for the shim

internal/cgrouputils/register.go
```diff
@@ -0,0 +1,35 @@
...
+const (
+	MemRoot string = "/sys/fs/cgroup/memory"
+)
+
+// RegisterMemoryThreshold registers an eventfd(2) and a memory threshold in
+// 'cgroup.event_control' of a cgroup. The eventfd will trigger when
+// memory.usage_in_bytes exceeds the threshold set.
+// https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt Section 9
+func RegisterMemoryThreshold(cgPath string, threshold int64) (uintptr, error) {
+	efd, err := newEventFD()
```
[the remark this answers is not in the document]
dcantah: Good idea that's much cleaner

service/gcs/main.go
dcantah: @jstarks I was going to log mem-usage, mem-usage-limit, mem-usage-max and the same for swap and kernel (might as well if we have the stats), anything else you can think of that would be good to include? We have the whole suite of stats available so everything in this https://github.com/containerd/cgroups/blob/6c3dec43a1030a55584ab8b5b181411029d43ecc/stats/v1/metrics.pb.go#L279

service/gcs/main.go
dcantah: I'll just defer the file closures in main instead, I agree it's odd

internal/cgrouputils/register.go
```diff
@@ -0,0 +1,39 @@
...
+// https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt Section 9
+func RegisterMemoryThreshold(cgPath string, threshold int64) (*os.File, error) {
+	efd, err := unix.Eventfd(0, unix.EFD_CLOEXEC)
+	if err != nil {
+		return nil, err
+	}
+	efdFile := os.NewFile(uintptr(efd), "efd")
+	f, err := os.Open(filepath.Join(cgPath, "memory.usage_in_bytes"))
+	if err != nil {
+		return nil, err
```
jstarks, line 29: `efdFile` is leaked here

jstarks: Maybe do a `defer func() { if err != nil { efdFile.Close() } }`?

service/gcs/main.go
```diff
@@ -5,21 +5,60 @@ import (
...
+	// Buffer must be >= 8 bytes for eventfd reads
+	// http://man7.org/linux/man-pages/man2/eventfd.2.html
+	buf := make([]byte, 8)
+	for {
+		_, err := efdFile.Read(buf)
+		if err != nil {
+			logrus.WithError(err).Warn("opengcs::readEvents - failed to read from eventfd")
+			return
+		}
+		msg := fmt.Sprintf("opengcs::readMemoryEvents - memory usage for cgroup '%s' exceeded '%d' byte threshold", cgName, threshold)
```
jstarks, line 50: Can we make this msg not dynamic and just incorporate the information into fields?

jstarks: (That way it's easier to parse it later.)

dcantah: The cgName and threshold can be passed into memoryLogFormat and moved to fields yes. The dynamic aspect of the message  carries through to the lines below this if we fail to retrieve the stats as we don't have anything to log besides the name and threshold though. So I guess there's two questions regarding if we fail to retrieve the stats 
1. Should I keep the appending of the "- failed to gather memory metrics"
2. If we fail to gather the metrics should we just log the message with cgName and threshold as fields or should we stay consistent with a success and log all of the fields but everything besides cgName and threshold be default values.

jstarks: @kevpar can you provide guidance

service/gcs/main.go
```diff
@@ -5,21 +5,60 @@ import (
...
+}
+
+func readMemoryEvents(efdFile *os.File, cgName string, threshold int64, cg cgroups.Cgroup) {
+	// Buffer must be >= 8 bytes for eventfd reads
+	// http://man7.org/linux/man-pages/man2/eventfd.2.html
+	buf := make([]byte, 8)
+	for {
+		_, err := efdFile.Read(buf)
+		if err != nil {
+			logrus.WithError(err).Warn("opengcs::readEvents - failed to read from eventfd")
```
kevpar, line 47: I don't think we should keep using this prefix like `opengcs::readEvents - ` on our log statements. It's something I want to start removing since it usually just seems to add noise.

service/gcs/main.go
kevpar: You can pre-populate a `logrus.Entry` with the initial fields:

```
msg := "monitored cgroup memory exceeded threshold"
entry := logrus.WithFields(logrus.Fields{
    "cgroup": cgName,
    "thresholdBytes": threshold,
})
time.Sleep(time.Second)
metrics, err := cg.Stat()
if err != nil {
    entry.WithError(err).Error(msg)
} else {
    logrus.WithFields(memoryLogFormat(metrics)).Warn(msg)
}
```

Also:

- If we fail to get the memory stats we should log that as an error, rather than a warning.
- We should log the error that we receive from `cg.Stat()`.
- I'm not sure we need to add to the message in the failure case, we can probably just log it at a different level and include the error as a field.

service/gcs/main.go
```diff
@@ -5,21 +5,60 @@ import (
...
+			return
+		}
+		msg := fmt.Sprintf("opengcs::readMemoryEvents - memory usage for cgroup '%s' exceeded '%d' byte threshold", cgName, threshold)
+		time.Sleep(time.Second)
```
kevpar, line 51: What happens if we don't sleep here? Are the values noticeably different?

internal/cgrouputils/register.go
```diff
@@ -0,0 +1,39 @@
...
+// Memory cgroup root path
+const (
+	MemRoot string = "/sys/fs/cgroup/memory"
+)
+
+// RegisterMemoryThreshold registers an eventfd(2) and a memory threshold in
+// 'cgroup.event_control' of a cgroup. The eventfd will trigger when
+// memory.usage_in_bytes exceeds the threshold set.
+// https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt Section 9
+func RegisterMemoryThreshold(cgPath string, threshold int64) (*os.File, error) {
```
kevpar, line 21: Recommend breaking the code up a bit into stanzas: One stanza for setting up `efd`, one for `f`, and then one that does the writing.

dcantah: Are you referring to different functions or just separation of logic indicated by newlines as below 
`go
efd, err := unix.Eventfd(0, unix.EFD_CLOEXEC)
	if err != nil {
		return nil, err
	}
	efdFile := os.NewFile(uintptr(efd), "efd")
	defer func() {
		if err != nil {
			efdFile.Close()
		}
	}()
        //newline added
	f, err := os.Open(filepath.Join(cgPath, "memory.usage_in_bytes"))
	if err != nil {
		return nil, err
	}
	defer f.Close()
	// newline added
	data := fmt.Sprintf("%d %d %d", efdFile.Fd(), f.Fd(), threshold)
	ecPath := filepath.Join(cgPath, "cgroup.event_control")
	if err := ioutil.WriteFile(ecPath, []byte(data), 0700); err != nil {
		efdFile.Close()
		return nil, err
	}
	return efdFile, err
`

dcantah: don't know why that one line is non indented lol

service/gcs/main.go
dcantah: For the gcs case yes as the event gets fired and theres no action taken besides logging so the numbers can continue to change. For the oom case not really.

service/gcs/main.go
dcantah: We should have a pr to do something similar in vain to https://github.com/microsoft/hcsshim/pull/791 this eventually as well then

service/gcs/main.go
kevpar: In that case, why do we sleep? Don't we want to know what these values are at the time we hit the threshold?

service/gcs/main.go
kevpar: I'm in favor of continuing to improve log statements as we come across them. We don't necessarily have to do them all at once.

internal/cgrouputils/register.go
kevpar: Just separate them with blank lines please.

dcantah: Ok cool. The comments were added just to make it more obvious, they weren't going to be kept if that wasn't clear

service/gcs/main.go
jstarks: I suggested this originally. My intuition was that it might be helpful to wait a second in case there are a series of allocations. I am guessing that the "max" fields will be the high water mark, in which case we don't have to worry about what the memory state is at the precise time we exceed the threshold.

dcantah: max is the highwater mark correct, usage is the snapshot at the current point in time. On event notification if this is the first time we've exceeded the threshold Usage and Max would be very similar (depending on how long Stat takes to run. Stat calls the corresponding Stat method on all subsystems not just memory).

internal/cgrouputils/register.go
```diff
@@ -0,0 +1,47 @@
...
+	efd, err := unix.Eventfd(0, unix.EFD_CLOEXEC)
+	if err != nil {
+		return nil, err
+	}
+
+	efdFile := os.NewFile(uintptr(efd), "efd")
+	defer func() {
+		if err != nil {
```
jstarks, line 29: I think this won't do what you want unless you name the error return value `err` above, e.g.

`func RegisterMemoryThreshold(cgPath string, threshold int64) (_ *os.File, err error) {`

dcantah: Are you sure? I made a dummy program to test this and it seems to function. err is already defined above

kevpar: This way can work in some cases, but it's not very robust, and simple refactorings can inadvertently cause the clean up code to stop working.

For instance, in the below example, clean up does not occur:

```
package main

import "fmt"

func main()        { fmt.Printf("error: %s\n", test()) }
func works() error { return nil }
func fails() error { return fmt.Errorf("failure") }

func test() error {
	err := works()
	if err != nil {
		return err
	}
	defer func() {
		if err != nil {
			fmt.Println("cleaning up")
		}
	}()
	if err := fails(); err != nil {
		return err
	}
	return nil
}
```

This is because now the scope of the second `err` declaration is in a nested block, so the `defer` doesn't see this value. Instead, it only sees the `err` value in the top-level scope. However, if you use a named return value, it will work correctly.

dcantah: This is what I was testing with which is pretty much this scenario but that makes sense. Lets change it to be safe. 
```go
func errorTest() (string, error) {
	err := fmt.Errorf("oh no an error")
	defer func(){
		if err != nil {
			fmt.Println("oh no again")
		}
	}()
	if err != nil {
		return "", err	
	}
	return "success", nil	
}```

File: internal/cgrouputils/register.go [hunk 11]
```diff
@@ -0,0 +1,47 @@
...
+	f, err := os.Open(filepath.Join(cgPath, "memory.usage_in_bytes"))
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	data := fmt.Sprintf("%d %d %d", efdFile.Fd(), f.Fd(), threshold)
+	ecPath := filepath.Join(cgPath, "cgroup.event_control")
+	if err := ioutil.WriteFile(ecPath, []byte(data), 0700); err != nil {
+		efdFile.Close()
```
kevpar, line 43: Don't want this `Close` call anymore.

service/gcs/main.go
```diff
@@ -16,11 +18,56 @@ import (
 	"github.com/Microsoft/opengcs/service/gcs/runtime/runc"
 	"github.com/Microsoft/opengcs/service/gcs/transport"
 	"github.com/containerd/cgroups"
+	v1 "github.com/containerd/cgroups/stats/v1"
```
kevpar, line 21: I think we should import this with a clearer name like `cgroupstats`.

service/gcs/main.go
```diff
@@ -16,11 +18,56 @@ import (
...
 	oci "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
 	"go.opencensus.io/trace"
 )
 
+func memoryLogFormat(metrics *v1.Metrics) logrus.Fields {
+	return logrus.Fields{
+		"memory-usage":       metrics.Memory.Usage.Usage,
```
kevpar, line 29: Our general convention for field names is to use `camelCase`.

service/gcs/main.go
kevpar: That seems reasonable. My request then is we add a comment here explaining why we are sleeping (i.e. in case there is a series of allocations).

service/gcs/main.go
```diff
@@ -156,6 +199,21 @@ func main() {
 		logrus.WithError(err).Fatal("opengcs::main - failed add gcs pid to gcs cgroup")
 	}
 
+	gefd, err := cgrouputils.RegisterMemoryThreshold(filepath.Join(cgrouputils.MemRoot, "gcs"), gcsLimit)
+	if err != nil {
+		logrus.WithError(err).Fatal("opengcs::main - failed to register memory threshold for gcs cgroup")
```
kevpar, line 204: Let's remove the `opengcs::main - ` prefix from these log statements.

dcantah:
https://github.com/containerd/cgroups/pull/154 @kevpar Will fix the remaining feedback but may want to wait till this gets merged now

service/gcs/main.go
dcantah: Want me to rid all of main of these or just this commits introduced ones?

kevpar: I'm fine with fixing things as you come across them. If you're doing something more broadly across the repo it would be better in a separate PR.

dcantah: I'll just add a separate commit to remove them in this pr as they're all in main, might as well I feel.

dcantah:
Cgroups pr has been merged. Making another commit to vendor this/move to new methods.

service/gcs/main.go
```diff
@@ -16,11 +16,57 @@ import (
...
+}
+
+func readMemoryEvents(efdFile *os.File, cgName string, threshold int64, cg cgroups.Cgroup) {
+	// Buffer must be >= 8 bytes for eventfd reads
+	// http://man7.org/linux/man-pages/man2/eventfd.2.html
+	buf := make([]byte, 8)
+	for {
+		_, err := efdFile.Read(buf)
+		if err != nil {
+			logrus.WithError(err).Warn("failed to read from eventfd")
```
kevpar, line 46: I would probably make this error level, since it's a clear indication that something has failed.

service/gcs/main.go
```diff
@@ -139,27 +185,40 @@ func main() {
...
+	defer gefdFile.Close()
+
+	oom, err := containersControl.OOMEventFD()
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to retrieve the container cgroups oom eventfd")
 	}
+	oomFile := os.NewFile(oom, "cefd")
+	defer oomFile.Close()
 
+	go readMemoryEvents(gefdFile, "gcs", int64(*gcsMemLimitBytes), gcsControl)
```
kevpar, line 216: I think we should make these cgroup names be the exact value, so `"/gcs"` instead of `"gcs"`, and `"/containers"` instead of `"containers"`.

service/gcs/main.go
```diff
@@ -16,11 +16,57 @@ import (
...
+		"kernelUsageLimit": metrics.Memory.Kernel.Limit,
+	}
+}
+
+func readMemoryEvents(efdFile *os.File, cgName string, threshold int64, cg cgroups.Cgroup) {
+	// Buffer must be >= 8 bytes for eventfd reads
+	// http://man7.org/linux/man-pages/man2/eventfd.2.html
+	buf := make([]byte, 8)
+	for {
+		_, err := efdFile.Read(buf)
```
kevpar, line 44: You can combine the `Read` and `if` into one line here.

service/gcs/main.go
kevpar: Also, we should log `cgName` here.

PR #373, opened 2020-03-24
Continuously log /dev/kmsg

dcantah:
* Added new internal package kmsg
* Continuously log all /dev/kmsg entries <= the priority level chosen
* Added new flag 'kmsgLogLevel' to be able to set the priority of entries we would
like to have logged (default is 4).

### DIFF
--- a/internal/kmsg/kmsg.go
+++ b/internal/kmsg/kmsg.go
@@ -1,0 +1,149 @@
+// Package kmsg contains support for parsing Linux kernel log entries read from
+// /dev/kmsg. These are the same log entries that can be read via the `dmesg`
+// command. Each read from /dev/kmsg is guaranteed to return a single log entry,
+// so no line-splitting is required.
+//
+// More information can be found here:
+// https://www.kernel.org/doc/Documentation/ABI/testing/dev-kmsg
+package kmsg
+
+import (
+	"errors"
+	"os"
+	"strconv"
+	"strings"
+	"syscall"
+
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	// ErrInvalidFormat indicates the kmsg entry failed to parse.
+	ErrInvalidFormat = errors.New("invalid kmsg format")
+)
+
+// LogLevel represents the severity/priority of a log entry in the kernels
+// ring buffer.
+// https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/kern_levels.h?id=HEAD
+type LogLevel uint8
+
+func (logLevel LogLevel) String() string {
+	return levels[logLevel]
+}
+
+const (
+	Emerg LogLevel = iota
+	Alert
+	Crit
+	Err
+	Warning
+	Notice
+	Info
+	Debug
+)
+
+var levels = [...]string{
+	"Emerg",
+	"Alert",
+	"Crit",
+	"Err",
+	"Warning",
+	"Notice",
+	"Info",
+	"Debug",
+}
+
+// Entry is a single log entry in kmsg.
+type Entry struct {
+	Priority           LogLevel
+	Facility           uint8
+	Seq                uint64
+	TimeSinceBootMicro uint64
+	Flags              string
+	Message            string
+}
+
+func (ke *Entry) logFormat() logrus.Fields {
+	return logrus.Fields{
+		"priority":           ke.Priority.String(),
+		"facility":           ke.Facility,
+		"seq":                ke.Seq,
+		"timesincebootmicro": ke.TimeSinceBootMicro,
+		"flags":              ke.Flags,
+		"message":            ke.Message,
+	}
+}
+
+// Parse takes a single kmsg log entry string and returns a struct representing
+// the components of the log entry.
+func parse(s string) (*Entry, error) {
+	fields := strings.SplitN(s, ";", 2)
+	if len(fields) < 2 {
+		return nil, ErrInvalidFormat
+	}
+	prefixFields := strings.SplitN(fields[0], ",", 5)
+	if len(prefixFields) < 4 {
+		return nil, ErrInvalidFormat
+	}
+	syslog, err := strconv.ParseUint(prefixFields[0], 10, 16)
+	if err != nil {
+		return nil, ErrInvalidFormat
+	}
+	seq, err := strconv.ParseUint(prefixFields[1], 10, 64)
+	if err != nil {
+		return nil, ErrInvalidFormat
+	}
+	timestamp, err := strconv.ParseUint(prefixFields[2], 10, 64)
+	if err != nil {
+		return nil, ErrInvalidFormat
+	}
+	return &Entry{
+		Priority:           LogLevel(syslog & 0x7),
+		Facility:           uint8(syslog >> 3),
+		Seq:                seq,
+		TimeSinceBootMicro: timestamp,
+		Flags:              prefixFields[3],
+		Message:            fields[1],
+	}, nil
+}
+
+// ReadForever reads from /dev/kmsg forever unless /dev/kmsg cannot be opened.
+// Every entry with priority <= 'logLevel' will be logged.
+func ReadForever(logLevel LogLevel) {
+	file, err := os.Open("/dev/kmsg")
+	if err != nil {
+		logrus.WithError(err).Error("failed to open /dev/kmsg")
+		return
+	}
+	defer file.Close()
+	// Reuse buffer for entries
+	// Buffer size from: https://elixir.bootlin.com/linux/latest/source/include/linux/printk.h#L44
+	buf := make([]byte, 8192)
+	for {
+		n, err := file.Read(buf)
+		if err != nil {
+			// "In case messages get overwritten in the circular buffer while
+			// the device is kept open, the next read() will return -EPIPE,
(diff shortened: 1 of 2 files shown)


dcantah:
This is essentially the same PR/feature add as https://github.com/microsoft/opengcs/pull/367 just with a few tweaks to be able to skip logging entries lower than a certain priority and some data regarding the number of logs per priority level. 

Running a trace on the cri-containerd test suite yields 
-Level 3: Around 3-4 logs 
-Level 4 (default): Around 9-12 logs
-Level 5: Around 40-50
-Level 6: 330+ and the majority of these are system initialization entries 
-Level 7: 360+

internal/kmsg/kmsg.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+var (
+	// ErrInvalidFormat indicates the kmsg entry failed to parse.
+	ErrInvalidFormat = errors.New("invalid kmsg format")
+)
+
+// Entry is a single log entry in kmsg.
+type Entry struct {
+	Priority           uint8
```
kevpar: It looks like kmsg uses the standard syslog priority levels. I think we should define a type with those values, and use that instead of a raw uint8. I would use the keywords here for the names: https://en.wikipedia.org/wiki/Syslog#Severity_level

kevpar: I mostly care about this now because we allow it as a filtering parameter, and it's a lot more meaningful to see that we are filtering to "<= info" rather than "<= 4".

dcantah: I agree. An enum would be easier to reason with

dcantah: @kevpar You think logging the priority name would be a good idea as well? I don't think it has to be a member of the struct but specifically referring to adding a String method to the LogLevel type I'm going to make and doing this

dcantah: @kevpar You think we should also log the priority name 'Warning', 'Emergency' etc with the other fields? I don't think it has to be a member of the struct but specifically referring to adding a String method to the **LogLevel** type I'm going to make and just adding it as a field on the **logrus.Fields** thats returned from **logFormat**. 

```go
// Entry is a single log entry in kmsg.
type Entry struct {
	Priority           LogLevel
	Facility           uint8
	Seq                uint64
	TimeSinceBootMicro uint64
	Flags              string
	Message            string
}

func (ke *Entry) logFormat() logrus.Fields {
	return logrus.Fields{
		"priority":           ke.Priority,
		"priority-name":      ke.Priority.String(),
		"facility":           ke.Facility,
		"seq":                ke.Seq,
		"timesincebootmicro": ke.TimeSinceBootMicro,
		"flags":              ke.Flags,
		"message":            ke.Message,
	}
}
```

kevpar: I'd be fine with logging the priority name instead of the integer, but I don't think it's worth logging both.

internal/kmsg/kmsg.go
```diff
@@ -0,0 +1,149 @@
...
+		"Warning",
+		"Notice",
+		"Informational",
+		"Debug",
+	}
+	return levels[logLevel]
+}
+
+const (
+	Emergency LogLevel = iota
```
kevpar: I think we should use the same names as in the kernel. e.g. `Emerg`, `Alert`, `Crit`, etc.

dcantah: Even when logging the name or just for the enum vars?

internal/kmsg/kmsg.go
```diff
@@ -0,0 +1,149 @@
...
+	ErrInvalidFormat = errors.New("invalid kmsg format")
+)
+
+// LogLevel represents the severity/priority of a log entry in the kernels
+// ring buffer.
+// https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/kern_levels.h?id=HEAD
+type LogLevel uint8
+
+func (logLevel LogLevel) String() string {
+	levels := [8]string{
```
kevpar: You can write this as `levels := [...]string{` instead.

internal/kmsg/kmsg.go
kevpar: I'd probably use the short name everywhere. Not a big deal though.

internal/kmsg/kmsg.go
```diff
@@ -0,0 +1,149 @@
...
+	ErrInvalidFormat = errors.New("invalid kmsg format")
+)
+
+// LogLevel represents the severity/priority of a log entry in the kernels
+// ring buffer.
+// https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/kern_levels.h?id=HEAD
+type LogLevel uint8
+
+func (logLevel LogLevel) String() string {
+	levels := [...]string{
```
jstarks: nit: this syntax is new to me. Why isn't this just `[]string`?

internal/kmsg/kmsg.go
```diff
@@ -0,0 +1,149 @@
...
+	buf := make([]byte, 8192)
+	for {
+		n, err := file.Read(buf)
+		if err != nil {
+			// "In case messages get overwritten in the circular buffer while
+			// the device is kept open, the next read() will return -EPIPE,
+			// and the seek position be updated to the next available record.
+			// Subsequent reads() will return available records again."
+			if err == syscall.EPIPE {
+				logrus.Info("kmsg entry overwritten; skipping entry")
```
jstarks: Should this be `Warn`?

internal/kmsg/kmsg.go
```diff
@@ -0,0 +1,149 @@
...
+				continue
+			}
+			logrus.WithError(err).Error("kmsg read failure")
```
jstarks: On read failure, should we terminate the loop? Otherwise it seems like we might get into an infinite loop.

internal/kmsg/kmsg.go
```diff
@@ -0,0 +1,149 @@
...
+		line := string(buf[:n])
+		entry, err := parse(line)
+		if err != nil {
+			logrus.WithFields(logrus.Fields{
+				logrus.ErrorKey: err,
+				"line":          line,
+			}).Error("failed to parse kmsg entry")
+		} else {
+			if entry.Priority > logLevel {
+				continue
```
jstarks: nit: the continue here is just skipping the one line below it, but my brain had to spend time to make sure it wasn't skipping anything else. It's probably clearer in this case to invert the if condition and put the log statement inside it.

internal/kmsg/kmsg.go
kevpar: `[]string` will produce a slice, whereas `[...]string` will produce an array sized according to the number of elements used to initialize it. In this case, I don't think it matters.

jstarks: It would be interesting to look at the assembly.

dcantah: If I know the number of elements ahead of time I normally make an array but it really doesn't matter in this case. I wasn't aware of this syntax either, it was previously [8]string so I learned something new :)

internal/kmsg/kmsg.go
dcantah: Yes good catch

internal/kmsg/kmsg.go
jstarks: I looked at the assembly using `go build -gcflags -S file.go`. At a quick glance, it doesn't seem to make much difference between slice and array. What does seem to make a difference is whether the slice/array is a local or a global. It looks like the Go compiler will copy the slice/array onto the function's stack when you declare it inside like this.

So probably it's better to move this out as a global.

kevpar: That's kind of annoying. I would have hoped we would get that optimization if Go sees it doesn't need to really be function local.

I poked around a little at godbolt.org. I don't understand a lot of the assembly, but it looks like the array method avoids a heap allocation (the slice method has a call to `runtime.newobject`).

jstarks: I didn't see runtime.newobject in any of the 4 cases (slice vs array, globla vs local) when I ran it locally on Go 1.13.4.

kevpar: Interesting, could just be a platform difference. I'm using the "gc x86 1.13" configuration, which is presumably building for Linux.

kevpar: https://godbolt.org/z/F9jMR3

I don't see a way to set env vars via godbolt, unfortunately, so I can't change `GOOS`.

internal/kmsg/kmsg.go
```diff
@@ -0,0 +1,150 @@
+// Package kmsg contains support for parsing Linux kernel log entries read from
+// /dev/kmsg. These are the same log entries that can be read via the `dmesg`
+// command. Each read from /dev/kmsg is guaranteed to return a single log entry,
+// so no line-splitting is required. If the read buffer is not large enough to
+// contain the log entry, `os.ErrInvalid` is returned.
```
jstarks, line 5: This doesn't seem true (anymore?)

internal/kmsg/kmsg.go
jstarks: Ah, I see. If you update the print calls to just grab one of the array/slice entries via indexing (a[i]) then you don't see the runtime.newobject call anymore.

PR #375, opened 2020-06-03
Remove dax mount option to avoid 5.4 kernel hard failure

dcantah:
* Removed dax mount option to circumvent the 5.4 kernel hard failure on
unsupported dax devices. Previously if the block device wasn't supported
the kernel would simply silently fail. This behavior was changed with this patch
https://patchwork.kernel.org/patch/10631361/.

This change should be reverted whenever dax is properly supported for our
pmem devices.

### DIFF
--- a/internal/storage/pmem/pmem.go
+++ b/internal/storage/pmem/pmem.go
@@ -26,7 +26,7 @@ var (
 // automatically cleaned up.
 //
 // Note: For now the platform only supports readonly pmem that is assumed to be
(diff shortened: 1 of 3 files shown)


kevpar:
I know we don't run them automatically, but it would be good to update the test here as well: https://github.com/microsoft/opengcs/blob/master/internal/storage/pmem/pmem_test.go

dcantah:
@kevpar true true. Fixing

PR #376, opened 2020-06-14
Fix pmem build error 

dcantah:
* Fix 'pmem declared and not used' build error introduced with https://github.com/microsoft/opengcs/pull/375.
When testing things to solve the 5.4 kernel build failure I had been doing all of
the work on a '54-kernel-failure' branch, so when we discovered the dax issue I
added the fix to this branch and built off of it to verify. When seeing that this
resolved the issue I created a new branch and just added the removing of dax changes
and forgot to add what is in this commit before pushing.


dcantah:
@microsoft/containerplat we should build the repo/run tests like we do for the shim to avoid things like this :frowning:

PR #378, opened 2020-08-22
Touchup OOM/memory limit logging events

dcantah:
* An event gets sent during cgroup teardown so avoid logging an event for this case.

* Fix returning nil stats if the LCOW kernel doesn't have CONFIG_MEMCG_SWAP
(no memsw cgroup entries).

* Add additional gcs start timestamp and current timestamp fields to logs.

### DIFF
--- a/service/gcs/main.go
+++ b/service/gcs/main.go
@@ -36,9 +36,10 @@ func memoryLogFormat(metrics *cgroupstats.Metrics) logrus.Fields {
 	}
 }
 
-func readMemoryEvents(efdFile *os.File, cgName string, threshold int64, cg cgroups.Cgroup) {
+func readMemoryEvents(startTime time.Time, efdFile *os.File, cgName string, threshold int64, cg cgroups.Cgroup) {
 	// Buffer must be >= 8 bytes for eventfd reads
 	// http://man7.org/linux/man-pages/man2/eventfd.2.html
+	count := 0
 	buf := make([]byte, 8)
 	for {
 		if _, err := efdFile.Read(buf); err != nil {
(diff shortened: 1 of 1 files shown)


service/gcs/main.go
```diff
@@ -39,13 +39,22 @@ func memoryLogFormat(metrics *cgroupstats.Metrics) logrus.Fields {
...
 	// http://man7.org/linux/man-pages/man2/eventfd.2.html
+	triggered := false
 	buf := make([]byte, 8)
 	for {
 		if _, err := efdFile.Read(buf); err != nil {
 			logrus.WithError(err).WithField("cgroup", cgName).Error("failed to read from eventfd")
 			return
 		}
 
+		_, err := os.Lstat(fmt.Sprintf("/sys/fs/cgroup/memory/%s/cgroup.event_control", cgName))
```
kevpar, line 50: What's the point of checking if this file exists? It's not clear from the comment below.

service/gcs/main.go
```diff
@@ -212,8 +221,8 @@ func main() {
 	oomFile := os.NewFile(oom, "cefd")
 	defer oomFile.Close()
 
-	go readMemoryEvents(gefdFile, "/gcs", int64(*gcsMemLimitBytes), gcsControl)
-	go readMemoryEvents(oomFile, "/containers", containersLimit, containersControl)
+	go readMemoryEvents(gefdFile, "gcs", int64(*gcsMemLimitBytes), gcsControl)
```
kevpar, line 224: Why are we changing the cgroup names here?

dcantah: This is just what gets printed out during logs. If it'd be more helpful to see "/gcs" vs. "gcs" then I can change back. It just looked better in the fmt.Sprintf call also haha

service/gcs/main.go
dcantah: https://github.com/opencontainers/runc/blob/e5f2eae5a56b89a801e1482b0d079051acca97f2/libcontainer/notify_linux.go#L55 Found this from runc. I don't think we're actually even seeing this event being sent as otherwise you'd think for N pods launched we'd have at minimum N events in our logs but thats not the case. Mostly just to be safe. Would you like it explained more?

service/gcs/main.go
```diff
@@ -39,13 +39,22 @@ func memoryLogFormat(metrics *cgroupstats.Metrics) logrus.Fields {
 func readMemoryEvents(efdFile *os.File, cgName string, threshold int64, cg cgroups.Cgroup) {
 	// Buffer must be >= 8 bytes for eventfd reads
 	// http://man7.org/linux/man-pages/man2/eventfd.2.html
+	triggered := false
```
kevpar, line 42: I'm not convinced we should be blocking subsequent triggers of the event. Won't there be cases where memory usage could fluctuate, and we want to get the event each time? What is the problem this is trying to solve?

service/gcs/main.go
kevpar: What I get from that is that you're checking if `cgroup.event_control` exists as a way to check if the cgroup is being torn down? I didn't feel that was explained by the comment below. I would move the comment up above the `Lstat` line, and rephrase it (e.g. `Sometimes an event is sent during cgroup teardown, but does not indicate that the threshold was actually crossed. In the teardown case the cgroup.event_control file won't exist anymore, so check that to determine if we should ignore this event.).

service/gcs/main.go
kevpar: I think we should keep `/gcs` as that matches the cgroup names used above.

dcantah: That's just the path we pass in to StaticPath, it could also be just "gcs" or "containers". I'll revert this though, I think the path is nicer after looking again

service/gcs/main.go
dcantah: Me and Howard were basically just trying to cut down on noise in the logs as it was hard to correlate to where this was actually impacting runc starting etc. I assumed we'd only really care about the first instance of this as the pods are generally short lived. If we want to log every time it goes over 50 I can remove it

service/gcs/main.go
dcantah: Changing, thats much better

service/gcs/main.go
kevpar: We don't know that pods are necessarily short lived. A pod could be kept running for days with different containers in it, for all we know. :)

kevpar: If knowing the first occurrence of the threshold being hit is helpful, let's just add a counter here and log that with the event info?

dcantah: No you're right, that was a dumb thought on my part. This was mostly to fix the short lived pod scenarios like functions etc where we're seeing an excess of these being fired for short lived pods but it'd be good to see for long lived scenarios the fluctuations.

dcantah:
@kevpar Sorry was waiting to test this on a kernel that has swap enabled. Just got that this morning so will test it out.

PR #380, opened 2020-09-17
Add new bridge call to update runc container

katiewasnothere:
This PR allows for LCOW container updates as part of the work to allow for dynamic resource changes during container runtime

### DIFF
--- a/service/gcs/prot/protocol.go
+++ b/service/gcs/prot/protocol.go
@@ -100,10 +100,12 @@ const (
 	ComputeSystemModifySettingsV1 = 0x10100a01
 	// ComputeSystemNegotiateProtocolV1 is the protocol negotiation request.
 	ComputeSystemNegotiateProtocolV1 = 0x10100b01
-	// ComputeSystemDumpStackV1 is the dump stack request
+	// ComputeSystemDumpStacksV1 is the dump stack request
 	ComputeSystemDumpStacksV1 = 0x10100c01
 	// ComputeSystemDeleteContainerStateV1 is the delete container request.
 	ComputeSystemDeleteContainerStateV1 = 0x10100d01
+	// ComputeSystemUpdateContainerV1 is the update container request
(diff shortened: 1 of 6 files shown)


service/gcs/prot/protocol.go
```diff
@@ -448,6 +453,12 @@ type ContainerGetProperties struct {
 	Query string
 }
 
+// ContainerUpdate is the message sent to update a container's resources
+type ContainerUpdate struct {
+	MessageBase
+	Resources string
```
kevpar, line 459: This looks like it's just [LinuxResources](https://pkg.go.dev/github.com/opencontainers/runtime-spec/specs-go#LinuxResources), should we use that type instead?

katiewasnothere: I was running into issues with unmarshalling the `ContainerUpdate` struct when doing that, when I googled for it, people mentioned it was because there was an embedded type that needed to be individually unmarshalled. In that case, I figured it would be easier to keep this a string since runc reads from stdin (so reads in a string) anyways. I can look into it again though to see if I was maybe just doing something wrong.

dcantah (approved):
LGTM

kevpar:
Two thoughts on this, but I don't have time to look into them further this week:
- Should we re-use the existing modify settings bridge message?
- Should we define a message struct that has something like `WindowsResources` and `LinuxResources` members, so we can have some stronger typing, but also use this for Windows in the future (without diverging the bridge protocol).

kevpar:
I don't think those items necessarily need to block, but would be good to have some understanding of them.

katiewasnothere:
> Two thoughts on this, but I don't have time to look into them further this week:
> 
> * Should we re-use the existing modify settings bridge message?
> * Should we define a message struct that has something like `WindowsResources` and `LinuxResources` members, so we can have some stronger typing, but also use this for Windows in the future (without diverging the bridge protocol).

I'm going to merge this for now but spend some time this week looking into your comments and make any additional PRs for that.

PR #381, opened 2020-10-03
Write 1 to memory.use_hierarchy to enable hierarchy support

dcantah:
`Created nested cgroup for controller "memory" which has incomplete hierarchy support. Nested cgroups may change behavior in the future.
cgroup: "memory" requires setting use_hierarchy to 1 on the root.`

After booting the UVM, this is present in dmesg. There's numerous threads on this over the years from docker, lxc and many others. This should enable the hierarchy support we (believe) we're using for our nested container cgroups so the resource consumptions are properly propagated upwards to their parent cgroup (/containers).`

* Enable hierarchy support so memory usage and limits can flow upwards to the parent cgroup like we'd expect for the /container and its nested cgroups. It's not necessary to set this on any nested cgroups, only the root.

### DIFF
--- a/service/gcs/main.go
+++ b/service/gcs/main.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


ambarve:
So the memory cgroup that we were using earlier for separation of containers and gcs was not working the way we expected it to? 

Otherwise this LGTM!

ambarve (approved):

dcantah:
@ambarve AFAIK yes. I'd only ever screwed around with the GCS' cgroup but I spotted this in dmesg and it got me thinking that how the containers cgroup is configured currently and what we're expecting to happen isn't whats actually happening. We set a limit on the containers cgroup itself to get triggered when it reaches a limit, and the nested container cgroups at /containers/<container_id> are supposed to propagate their memory usage upwards to the /containers cgroup. Without this change memory.usage_in_bytes is always 0 so this limit is essentially useless to my knowledge.

PR #382, opened 2020-10-15
Revert previous change that issues a kill all on container stop for LCOW

katiewasnothere:
Reverts https://github.com/microsoft/opengcs/pull/329 to better match the behavior of other shims and reflect the users request more accurately. This avoids an issue where child processes are killed before getting the chance to be cleaned up gracefully. 


katiewasnothere:
Verified by running the full cri-containerd test suite and double checking success with https://github.com/microsoft/hcsshim/pull/778.

katiewasnothere:
@microsoft/containerplat  PTAL :)

dcantah (approved):
If we validated that this fixes the issue at hand (I know this matches other shims so no worries there) then LGTM. I believe you said you'd follow up with them about one last thing yesterday.

PR #383, opened 2020-10-21
Support for making multiple overlays in the same scratch path

dcantah:
* The work and upper dirs were hardcoded previously so if you had tried to create more than one overlay in a mountpoint (/dev/sda / container scratch volume) it would fail. This change just adds a simple check to see if workdir and upperdir
already exist and will create a unique name for them with a generated ID. This is the same method cri uses to generate container IDs.

### DIFF
--- a/internal/storage/overlay/overlay.go
+++ b/internal/storage/overlay/overlay.go
@@ -53,6 +53,11 @@ func Mount(ctx context.Context, layerPaths []string, upperdirPath, workdirPath, 
 
 	options := []string{"lowerdir=" + lowerdir}
 	if upperdirPath != "" {
+		if _, err := os.Stat(upperdirPath); err == nil {
+			// Upperdir already exists. This means we're using the scratch space from
+			// another container. Create a new upper dir with a unique name for this container.
+			upperdirPath += "-" + generateID()
+		}
 		if err := osMkdirAll(upperdirPath, 0755); err != nil {
 			return errors.Wrap(err, "failed to create upper directory in scratch space")
 		}
@@ -64,6 +69,11 @@ func Mount(ctx context.Context, layerPaths []string, upperdirPath, workdirPath, 
 		options = append(options, "upperdir="+upperdirPath)
 	}
 	if workdirPath != "" {
+		if _, err := os.Stat(workdirPath); err == nil {
+			// Workdir already exists. This means we're using the scratch space from
+			// another container. Create a new work dir with a unique name for this container.
+			workdirPath += "-" + generateID()
+		}
 		if err := osMkdirAll(workdirPath, 0755); err != nil {
 			return errors.Wrap(err, "failed to create workdir in scratch space")
 		}
--- a/internal/storage/overlay/util.go
+++ b/internal/storage/overlay/util.go
@@ -1,0 +1,13 @@
+package overlay
+
+import (
+	"crypto/rand"
+	"encoding/hex"
+)
+
+// generateID generates a random unique id.
+func generateID() string {
+	b := make([]byte, 32)
+	rand.Read(b)
+	return hex.EncodeToString(b)
+}


internal/storage/overlay/overlay.go
```diff
@@ -53,6 +53,11 @@ func Mount(ctx context.Context, layerPaths []string, upperdirPath, workdirPath,
 
 	options := []string{"lowerdir=" + lowerdir}
 	if upperdirPath != "" {
+		if _, err := os.Stat(upperdirPath); err == nil {
+			// Upperdir already exists. This means we're likely using the sandbox containers
+			// scratch as the containers scratch so we need to make a new name for the upperdir.
```
katiewasnothere, line 58: I feel like we don't really need to specifically call out that we're using the sanbox container's scratch here. This is pretty generic so this comment should just say:

```
Upperdir already exists. This means we are reusing a scratch space from another container. Create a new dir with a random name to use for this container's upperdir.
```

internal/storage/overlay/overlay.go
```diff
@@ -64,6 +69,11 @@ func Mount(ctx context.Context, layerPaths []string, upperdirPath, workdirPath,
 		options = append(options, "upperdir="+upperdirPath)
 	}
 	if workdirPath != "" {
+		if _, err := os.Stat(workdirPath); err == nil {
+			// Workdir already exists. This means we're likely using the sandbox containers
+			// scratch as the containers scratch so we need to make a new name for the workdir.
```
katiewasnothere, line 74: Same as above

katiewasnothere (approved):
LGTM

kevpar:
This is basically so that two callers can call `overlay.Mount` for different containers, but pass the same scratch path? It seems like it would be better for those callers to just ensure they are passing a unique path. Doing this by generating a new random path seems gross, and also makes it harder to track which directories are for which containers, if we need to debug something.

anmaxvl (approved):
lgtm. nit: not sure if making a utility function is worth it (to dedupe)

kevpar (changes requested):
Have concerns with this design (see my other comment)

dcantah:
> This is basically so that two callers can call `overlay.Mount` for different containers, but pass the same scratch path? It seems like it would be better for those callers to just ensure they are passing a unique path. Doing this by generating a new random path seems gross, and also makes it harder to track which directories are for which containers, if we need to debug something.

Correct, but by unique path do you mean where in the scratch space we would setup the upper/work dirs? Obviously we don't have anything in the CombinedLayers call to specify what we'd like the upper and work dirs to be named so I assume you mean specify a unique scratch path (Maybe just the path to the containers scratch space we're sharing/<container_id> would be good). We could verify this in the shim if the refcount on the scratch scsi disk is > 1, and then we can set the scratch path to whatever format we devise. Thoughts? @kevpar

dcantah:
@kevpar For example, instead of this currently:
```
/run/gcs/c/bbe16b243025f94faae0b5c9a1cbd387a2a302b1460abec87806f308f4b1efe2 # ls  
config.json
hostname
hosts
lost+found
resolv.conf
rootfs
runc.log
sandboxMounts
upper
upper-7ab34e3c6bf1ac7024a8b7293884c3ddc28f87f6dccbf7098bf0fe46d5d5d5ff
upper-c84ecf1b8d55c0180dd2892870428fbffb0a4461a70e4363002b3e902416588a
work
work-b8dbcd37da754299807d724594f3797e8b557261ea990d8afe0513c8a0b6f2a0
work-e8ac3f60b0034ec5620721495000834e05032be55e30c1dc7eb6cbfda215ec6a
```

We could have:
```
/run/gcs/c/bbe16b243025f94faae0b5c9a1cbd387a2a302b1460abec87806f308f4b1efe2 # ls  
config.json
hostname
hosts
lost+found
resolv.conf
rootfs
runc.log
sandboxMounts
upper
7ab34e3c6bf1ac7024a8b7293884c3ddc28f87f6dccbf7098bf0fe46d5d5d5ff <- contains upper and work dirs for 7ab34...
c84ecf1b8d55c0180dd2892870428fbffb0a4461a70e4363002b3e902416588a <- contains upper and work dirs for c84ecf...
work
```

kevpar:
Yes, I think the caller of the combine layers modify operation should be responsible for passing a correct directory to use. The `overlay` package shouldn't have to worry about being given a path that's already been used.

As for what is the correct place to determine this, and what the paths should be, I'm not sure. As this work to re-use a scratch VHD for several containers spans several components, it might be helpful to write up an issue in `Microsoft/hcsshim` with your proposed flow. That would make it easier to reason about the end-to-end.

dcantah:
@kevpar https://github.com/microsoft/hcsshim/issues/891 :)

PR #385, opened 2020-12-02
Cleanup process directory after process exits.

ambarve:
Whenever a new process is started inside a container we create a temporary directory for
that process at path `/run/gcsrunc/<container-id>/<pid>` to store some process related
information. However, the v2 workflow of execProcess didn't cleanup these directories when
the process exited. This can cause the tmpfs mounted at the /run to get full for long
running containers. This change adds the change for cleaning up the directories after
process exits.

### DIFF
--- a/internal/runtime/hcsv2/process.go
+++ b/internal/runtime/hcsv2/process.go
@@ -103,6 +103,13 @@ func newProcess(c *Container, spec *oci.Process, process runtime.Process, pid ui
 		// least one waiter has read the result
 		go func() {
 			p.writersWg.Wait()
+			// cleanup the process state
+			if derr := p.process.Delete(); derr != nil {
+				log.G(ctx).WithFields(logrus.Fields{
+					"cid": p.cid,
+					"pid": p.pid,
+				}).Debugf("process cleanup error: %s", derr)
+			}
 			c.processesMutex.Lock()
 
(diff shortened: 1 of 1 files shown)


internal/runtime/hcsv2/process.go
```diff
@@ -103,6 +103,13 @@ func newProcess(c *Container, spec *oci.Process, process runtime.Process, pid ui
 		// least one waiter has read the result
 		go func() {
 			p.writersWg.Wait()
+			// cleanup the process state
```
dcantah, line 106: Just to make sure I'm following from a quick skim, we'll essentially fall into here when the process is finished but the problem is we can't wait to clean up the process state at container stop time as this issue is for long running containers that might be spawning a bunch of processes during it's lifetime?

ambarve: That's correct. All these directories will automatically get cleaned up when the container stops but for long running containers that doesn't work.

dcantah: Gotcha ok 👍

dcantah (approved):
LGTM. I assume you ran some validation for this?

ambarve:
Yeah. Right now if you start a container and run some processes inside that container you will see directories with the name as the pid of those processes inside `run/gcsrunc/<container-id>`. With this change those directories will immediately be cleaned up when the process exits. I verified by rebuilding gcs.

PR #386, opened 2020-12-03
Allow passing any uid for container processes.

ambarve:
Usually if a username is provided when starting a process inside the container we look inside the /etc/passwd file of the container to find the uid and gid for that user. However, if a uid is provided instead of a username there is no need to look into the /etc/passwd file to see if that user exists.

This is in accordance with how things are handled in linux as can be seen here: https://github.com/opencontainers/runc/blob/master/libcontainer/user/user.go#L291 (this is called by docker).
In fact in the same function if the user string is passed as `uid:gid` (switch case 2) then we don't look up the uid inside the `/etc/passwd` file.

### DIFF
--- a/internal/runtime/hcsv2/spec.go
+++ b/internal/runtime/hcsv2/spec.go
@@ -46,10 +46,12 @@ func setProcess(spec *oci.Spec) {
 	}
 }
 
-// setUserStr sets `spec.Process` to the valid `userstr` based on the OCI Image
-// Spec v1.0.0 `userstr`.
+// setUserStr sets `spec.Process` to the valid `userstr` based on the OCI Image Spec
+// v1.0.0 `userstr`.
 //
-// Valid values are: user, uid, user:group, uid:gid, uid:group, user:gid
+// Valid values are: user, uid, user:group, uid:gid, uid:group, user:gid.
+// If uid is provided instead of the username then that value is not checked against the
+// /etc/passwd file to verify if the user with given uid actually exists.
 func setUserStr(spec *oci.Spec, userstr string) error {
 	setProcess(spec)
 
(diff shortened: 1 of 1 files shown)


anmaxvl (approved):
LGTM!

kevpar:
Would be good to check that a more typical Linux runtime (e.g. moby or containerd on Linux) works when you specify a UID that doesn't exist.

internal/runtime/hcsv2/spec.go
```diff
@@ -61,7 +63,8 @@ func setUserStr(spec *oci.Spec, userstr string) error {
 			// evaluate username to uid/gid
 			return setUsername(spec, userstr)
 		}
-		return setUserID(spec, int(v))
+		spec.Process.User.UID, spec.Process.User.GID = uint32(v), 0
```
jstarks, line 66: Hmm. This means we won't get the gid even if the user is listed in the passwd file. Does that match what happens on Linux?

jstarks: It looks like containerd always tries to parse.

ambarve:
@kevpar Yeah actually it is very easy to test that. On a linux machine (or on docker running over WSL) you can do `docker run -it ubuntu:20.04` and then exec inside that container with `docker exec -u 12345 -it <cid> /bin/bash`. This works without any issues. And if you run `id -u` on that prompt it shows that the user id is `12345`.

internal/runtime/hcsv2/spec.go
ambarve: Good catch. I will update this to first check if the user exists in /etc/passwd and use the gid in that case otherwise simply use 0 as the gid.

ambarve: Fixed.

anmaxvl (approved):

dcantah (approved):

PR #387, opened 2021-01-06
Ken/jan4

KenGordon:
Have the layers mounted by pmem Mount use dm-verity.
Note. Using this requires an lcow linux kernel and root fs builds with dm-verity, loop and dm-encrypt support.

### DIFF
--- a/.github/pull.yml
+++ b/.github/pull.yml
@@ -1,0 +1,5 @@
+version: "1"
+rules:
+  - base: master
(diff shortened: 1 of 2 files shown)


KenGordon:
Drop this as it targets the wrong branch.

KenGordon:
> Drop this as it targets the wrong branch.

Wrong fork indeed.

PR #388, opened 2021-01-13
Add a new generic hook for use in device setup

katiewasnothere:
This PR removes the `nvidiaPrestartHook` binary for assigning nvidia devices into containers in favor of `genericHook`. genericHook can take an arbitrary binary and arguments to execute as an oci-hook (can be any type of hook). genericHook has the ability to fill in templated options in the arguments given, which allows for filling in values that are not known until runtime of the hook. 

This PR additionally adds a function to setup an oci prestart hook to run ldconfig, which is needed particularly for device assignment to ensure drivers are loaded. 


service/gcsutils/gcstools/genericeHook.go
```diff
@@ -0,0 +1,144 @@
+package main
```
anmaxvl, line 1: typo in filename?

internal/runtime/hcsv2/workload_container.go
```diff
@@ -171,6 +176,22 @@ func setupWorkloadContainerSpec(ctx context.Context, sbid, id string, spec *oci.
 	return nil
 }
 
+// Helper function to create an oci prestart hook to run ldconfig
+func addLDConfigHook(ctx context.Context, spec *oci.Spec, args, env []string) error {
```
anmaxvl, line 180: doesn't look like this can return an error. is it just to have similar behavior as other hooks?

internal/runtime/hcsv2/nvidia_utils.go
```diff
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/Microsoft/opengcs/internal/storage/pci"
+	generichook "github.com/Microsoft/opengcs/service/gcsutils/genericHook"
```
kevpar, line 13: `genericHook` -- generally Go idiom is to not have capitals in package names.

service/gcsutils/gcstools/genericHook.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func genericHookMain() {
+	if err := runGenericHook(); err != nil {
+		logrus.Errorf("error in generic hook: %s", err)
+		debugFileToRead := os.Getenv(generichook.LogDebugFileEnvKey)
+		if debugFileToRead != "" {
+			logDebugFile(debugFileToRead)
+		}
+		os.Exit(-1)
```
dcantah, line 84: I assume this is supposed to be 1

anmaxvl (approved):

service/gcsutils/gcstools/genericHook.go
```diff
@@ -0,0 +1,144 @@
...
+func newTemplateContext(state *specs.State) *templateContext {
+	t := &templateContext{
+		state: state,
+	}
+	t.funcs = template.FuncMap{
+		"id":         t.id,
+		"pid":        t.pid,
+		"annotation": t.annotation,
+	}
+	return t
```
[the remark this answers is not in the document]
katiewasnothere: I'm not sure if that will work since we're referencing the struct's functions in the funcmap and if we get rid of `t` I'm not sure what happens to those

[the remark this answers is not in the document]
dcantah: Oh I'm dumb, carry on. Didn't even realize these were assigning from t (and also functions)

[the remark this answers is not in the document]
dcantah: Resolving!

internal/runtime/hcsv2/nvidia_utils.go
```diff
@@ -23,22 +24,32 @@ const lcowNvidiaMountPath = "/run/nvidia"
...
 
+const nvidiaToolBinary = "nvidia-container-cli"
+
 // TODO katiewasnothere: prestart hooks will be depracated, this needs to be moved to a createRuntime hook
 // described here: https://github.com/opencontainers/runtime-spec/blob/39c287c415bf86fb5b7506528d471db5405f8ca8/config.md#posix-platform-hooks
 // addNvidiaDevicePreHook builds the arguments for nvidia-container-cli and creates the prestart hook
 func addNvidiaDevicePreHook(ctx context.Context, spec *oci.Spec) error {
-	nvidiaToolBinary := "nvidiaPrestartHook"
-	nvidiaToolPath, err := exec.LookPath(nvidiaToolBinary)
+	genericHookBinary := "genericHook"
```
anmaxvl, line 33: is this supposed to be lower case as well?

dcantah (dismissed):

internal/runtime/hcsv2/nvidia_utils.go
katiewasnothere: Apparently that file didn't save when I edited it lol fixed

anmaxvl (approved):

PR #389, opened 2021-01-25
enable device-mapper for mounting container layers

anmaxvl:
Use device mapper to create and mount linear block devices that correspond to container layers
that were mapped onto a single VPMEM device.
The device offset and size are expected to be in bytes and properly aligned

For more context:
https://github.com/microsoft/hcsshim/pull/930 enables mapping multiple LCOW layers
onto a single VPMEM device.

Signed-off-by: Maksim An <maksiman@microsoft.com>

### DIFF
--- a/internal/storage/pmem/pmem.go
+++ b/internal/storage/pmem/pmem.go
@@ -8,6 +8,8 @@ import (
 	"os"
 
 	"github.com/Microsoft/opengcs/internal/oc"
+	"github.com/Microsoft/opengcs/internal/storage"
+	dm "github.com/Microsoft/opengcs/internal/storage/devicemapper"
 	"github.com/pkg/errors"
 	"go.opencensus.io/trace"
 	"golang.org/x/sys/unix"
@@ -19,6 +21,28 @@ var (
 	osRemoveAll = os.RemoveAll
 	unixMount   = unix.Mount
 )
+
+const (
+	pMemFmt         = "/dev/pmem%d"
+	linearDeviceFmt = "dm-linear-pmem%d-%d-%d"
+)
+
+func mountInternal(source, target string) (err error) {
+	if err := osMkdirAll(target, 0700); err != nil {
+		return err
+	}
+	defer func() {
+		if err != nil {
+			osRemoveAll(target)
+		}
+	}()
+
+	flags := uintptr(unix.MS_RDONLY)
+	if err := unixMount(source, target, "ext4", flags, "noload"); err != nil {
+		return errors.Wrapf(err, "failed to mount %s onto %s", source, target)
+	}
+	return nil
+}
 
 // Mount mounts the pmem device at `/dev/pmem<device>` to `target`.
 //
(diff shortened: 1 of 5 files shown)


internal/storage/pmem/pmem.go
```diff
@@ -8,16 +8,20 @@ import (
...
 // Test dependencies
 var (
-	osMkdirAll  = os.MkdirAll
-	osRemoveAll = os.RemoveAll
-	unixMount   = unix.Mount
+	osMkdirAll      = os.MkdirAll
+	osRemoveAll     = os.RemoveAll
+	unixMount       = unix.Mount
+	pMemFmt         = "/dev/pmem%d"
+	linearDeviceFmt = "dm-linear-pmem%d-%d-%d"
```
katiewasnothere, line 24: Could you move these two to a **const** block? The values in this block right now are used for mocking functions in our tests.

internal/storage/pmem/pmem.go
```diff
@@ -44,10 +48,84 @@ func Mount(ctx context.Context, device uint32, target string) (err error) {
...
+	dmpath, err = dm.CreateDevice(linearDeviceName, dm.CreateReadOnly, []dm.Target{lt})
+	if err != nil {
+		return errors.Wrapf(err, "failed to create dm-linear target: pmem device: %d, offset: %d", device, deviceStart)
+	}
+
+	defer func() {
+		if err != nil {
+			dm.RemoveDevice(linearDeviceName)
+		}
+	}()
```
katiewasnothere, line 93: I think you could just do all of this and then call `Mount` above, right?

service/gcs/prot/protocol.go
```diff
@@ -848,11 +848,20 @@ type MappedDirectoryV2 struct {
...
+	MountPath           string `json:",omitempty"`
+}
+
 // MappedVPMemDeviceV2 represents a VPMem device that is mapped into a guest
 // path in the V2 schema.
 type MappedVPMemDeviceV2 struct {
 	DeviceNumber uint32 `json:",omitempty"`
 	MountPath    string `json:",omitempty"`
+	// Mapping is ignored when MountPath isn't empty
+	Mapping VirtualPMemMapping `json:",omitempty"`
```
katiewasnothere, line 864: Feels weird to me to say that MappedVPMemDeviceV2's `MountPath` is empty when using DM but then the VirtualPMemMapping struct contains a field for `MountPath`. Could we instead remove the duplicated `MountPath` in VirtualPMemMapping and just check if `Mapping` is emtpy to know when to use DM?

internal/storage/pmem/pmem.go
```diff
@@ -44,10 +48,84 @@ func Mount(ctx context.Context, device uint32, target string) (err error) {
...
+
+	linearDeviceName := fmt.Sprintf(linearDeviceFmt, deviceNumber, deviceStart, deviceSize)
+	if err := dm.RemoveDevice(linearDeviceName); err != nil {
+		return errors.Wrapf(err, "failed to remove dm-linear target %s", linearDeviceName)
+	}
+	return nil
+}
+
+// LinearTargetPMem returns VPMEM specific dm-linear target
+func LinearTargetPMem(deviceNumber uint32, lengthInBytes, startInBytes int64) dm.Target {
```
katiewasnothere, line 126: does this need to be exported?

internal/storage/pmem/pmem.go
```diff
@@ -44,10 +48,84 @@ func Mount(ctx context.Context, device uint32, target string) (err error) {
...
+	startBlock := startInBytes / dm.BlockSize
+	sizeInBlocks := lengthInBytes / dm.BlockSize
+	devicePath := fmt.Sprintf(pMemFmt, deviceNumber)
```
katiewasnothere, line 129: I think it makes more sense to have the caller of this function pass in the `devicePath` to use rather than construct one here with a constant fmt. And in that case, I think this function would make more sense in the `devicemapper` package than here.

internal/storage/pmem/pmem.go
anmaxvl: yeah, good point. we could have something like `mountInternal(ctx context.Context, source, target string) {...}` and have `Mount` do `mountInternal(ctx, fmt.Sprintf(pmemfmt, device), target)`

internal/storage/pmem/pmem.go
anmaxvl: probably not 🤔

internal/storage/pmem/pmem.go
anmaxvl: or maybe we don't need it at all. I just wanted to hide the block calculations

internal/storage/pmem/pmem.go
```diff
@@ -36,6 +60,22 @@ func Mount(ctx context.Context, device uint32, target string) (err error) {
...
+	defer span.End()
+	defer func() { oc.SetSpanStatus(span, err) }()
+
+	span.AddAttributes(
+		trace.Int64Attribute("device", int64(device)),
+		trace.Int64Attribute("deviceStart", deviceStart),
+		trace.Int64Attribute("sectorSize", deviceSize),
+		trace.StringAttribute("target", target))
+
 	if err := osMkdirAll(target, 0700); err != nil {
```
katiewasnothere, line 79: This is handled in the new `mountInternal` func right?

anmaxvl: oh lol, missed this one.

katiewasnothere (dismissed):
LGTM :)

dcantah (dismissed):

internal/storage/pmem/pmem.go
```diff
@@ -36,18 +60,54 @@ func Mount(ctx context.Context, device uint32, target string) (err error) {
...
+	linearDeviceName := fmt.Sprintf(linearDeviceFmt, device, deviceStart, deviceSize)
+
+	span.AddAttributes(
+		trace.Int64Attribute("device", int64(device)),
+		trace.Int64Attribute("deviceStart", deviceStart),
+		trace.Int64Attribute("sectorSize", deviceSize),
+		trace.StringAttribute("target", target),
+		trace.StringAttribute("table", fmt.Sprintf("%s: '%d %d %s'", linearDeviceName, lt.SectorStart, lt.Length, lt.Params)))
+
+	var dmpath string
```
dcantah, line 83: Does this actually need to be declared, as we just create it one line down anyways? There's no path where it gets assigned to something in a conditional

dcantah (approved):

PR #390, opened 2021-01-26
Remove the v1 gcs

katiewasnothere:
LCOW is only supported on RS5+ which already had LCOW v2. Remove unused V1 code. 


service/gcs/bridge/bridge.go
```diff
@@ -198,28 +192,15 @@ type Bridge struct {
...
 // events to be listen for and intercepted on `mux` before forwarding
 // to `gcs` for handling.
-func (b *Bridge) AssignHandlers(mux *Mux, gcs core.Core, host *hcsv2.Host) {
-	b.coreint = gcs
+func (b *Bridge) AssignHandlers(mux *Mux, host *hcsv2.Host) {
 	b.hostState = host
 
 	// These are PvInvalid because they will be called previous to any protocol
 	// negotiation so they respond only when the protocols are not known.
 	if b.EnableV4 {
```
dcantah, line 200: Is there ever a time where we wouldn't use v4 going forward now?

dcantah:
A sea of red 😄

dcantah (approved):

anmaxvl (approved):

service/gcs/bridge/bridge.go
katiewasnothere: I suppose not, but I think I'll clean that up in a separate PR just to avoid any accidental bugs here.

PR #391, opened 2021-01-26
Change the update container bridge call to use the modify call instead

katiewasnothere:
This PR updates OpenGCS to reuse the `modify` bridge call when issuing a request to update a container. Since the modify call was previously only used if the requested containerID is that of the UVM's pause container (null container ID), this PR additionally updates the code path to allow requests with specific container ID's to issue update calls but prevents them from issuing any host specific settings changes (such as adding a new mapped device in or a VPCI device). 

### DIFF
--- a/internal/runtime/hcsv2/container.go
+++ b/internal/runtime/hcsv2/container.go
@@ -129,7 +129,7 @@ func (c *Container) Delete(ctx context.Context) error {
 	return c.container.Delete()
 }
 
-func (c *Container) Update(ctx context.Context, resources string) error {
+func (c *Container) Update(ctx context.Context, resources interface{}) error {
 	return c.container.Update(resources)
 }
 
@@ -171,4 +171,8 @@ func (c *Container) GetStats(ctx context.Context) (*v1.Metrics, error) {
 	}
 
 	return cg.Stat(cgroups.IgnoreNotExist)
+}
+
+func (c *Container) modifyContainerConstraints(ctx context.Context, rt prot.ModifyRequestType, cc *prot.ContainerConstraintsV2) (err error) {
+	return c.Update(ctx, cc.Linux)
 }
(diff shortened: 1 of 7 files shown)


katiewasnothere:
@kevpar FYI :)

internal/runtime/hcsv2/container.go
```diff
@@ -172,3 +172,10 @@ func (c *Container) GetStats(ctx context.Context) (*v1.Metrics, error) {
 
 	return cg.Stat(cgroups.IgnoreNotExist)
 }
+
+func (c *Container) modifyContainerConstraints(ctx context.Context, rt prot.ModifyRequestType, cc *prot.ContainerConstraintsV2) (err error) {
+	if c.id == "" {
+		return errors.New("invalid parameter, containerID cannot be empty")
```
anmaxvl, line 178: can this actually happen?

anmaxvl (dismissed):
LGTM

internal/runtime/hcsv2/container.go
katiewasnothere: Good point. I'll remove :)

internal/runtime/hcsv2/uvm.go
```diff
@@ -207,9 +207,36 @@ func (h *Host) ModifyHostSettings(ctx context.Context, settings *prot.ModifySett
...
+	if err != nil {
+		return err
+	}
+
+	switch settings.ResourceType {
+	case prot.MrtContainerConstraints:
+		return c.modifyContainerConstraints(ctx, settings.RequestType, settings.Settings.(*prot.ContainerConstraintsV2))
 	default:
-		return errors.Errorf("the ResourceType \"%s\" is not supported", settings.ResourceType)
+		return errors.Errorf("the ResourceType \"%s\" is not supported for containers", settings.ResourceType)
```
dcantah, line 231: Could use %q and get rid of the quotes

dcantah (dismissed):

anmaxvl (dismissed):

katiewasnothere:
@dcantah and/or @anmaxvl would you PTAL when you get the chance? :) Rebased to account for the recent V1 removal

anmaxvl (approved):

PR #392, opened 2021-02-16
Add cifs mount and loopback device modify calls

dcantah:
To facilitate teleport enabled images from being used for LCOW, add two new modify____
branches for ModifyHostSettings.

* Add cifs mounting logic and associated calls.

* Add loopback device mounting logic and associated calls.

* Add `DoInNetNS` to be able to execute a function inside a specific network namespace. This is
needed as the common LCOW scenario has the only available network adapter not exposed to the
default network namespace. Therefore, to carry out a cifs mount for example, we need to join
this network namespace, perform the action, and then leave on the way out.

### DIFF
--- a/internal/storage/cifs/cifs.go
+++ b/internal/storage/cifs/cifs.go
@@ -1,0 +1,102 @@
+package cifs
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+	"unicode"
+
+	"github.com/Microsoft/opengcs/internal/network"
+	"github.com/Microsoft/opengcs/internal/oc"
+	"github.com/pkg/errors"
+	"github.com/vishvananda/netns"
+	"go.opencensus.io/trace"
+	"golang.org/x/sys/unix"
+)
+
+const cifsMountOptions = "cache=loose,actimeo=10000,dir_mode=0777,file_mode=0777,sec=ntlmssp,mfsymlinks,ro,async"
+
+var (
+	osMkdirAll  = os.MkdirAll
+	osRemoveAll = os.RemoveAll
+	unixMount   = unix.Mount
+)
+
+// Mount creates a cifs mount at `target`. If `target` doesn't exist this call
+// will try and create it.
+//
+// The UVM generally doesn't have a network adapter in the default network namespace so this function
+// will lock the current goroutines thread, join a network namespace with an available adapter and then
+// perform the mount.
+func Mount(ctx context.Context, source, target, username, password string) (err error) {
+	_, span := trace.StartSpan(ctx, "cifs.Mount")
+	defer span.End()
+	defer func() { oc.SetSpanStatus(span, err) }()
+
+	span.AddAttributes(
+		trace.StringAttribute("target", target),
+		trace.StringAttribute("source", source))
+
+	if err := osMkdirAll(target, 0700); err != nil {
+		return err
+	}
+	defer func() {
+		if err != nil {
+			osRemoveAll(target)
+		}
+	}()
+
+	cmd := exec.Command(
+		"mount",
+		"-t",
+		"cifs",
+		source,
+		target,
+		"-o",
+		fmt.Sprintf("vers=3.0,username=%s,password=%s,%s", username, password, cifsMountOptions),
+	)
+
+	pid, err := getNSPid()
+	if err != nil {
+		return errors.Wrap(err, "failed to get network namespace PID to perform cifs mount")
+	}
+
+	// Get a reference to the new network namespace
+	ns, err := netns.GetFromPid(pid)
+	if err != nil {
+		return errors.Wrapf(err, "netns.GetFromPid(%d) failed", pid)
+	}
+	defer ns.Close()
+
+	return network.DoInNetNS(ns, cmd.Run)
+}
+
+// This is gross, there HAS to be a better way
+func getNSPid() (int, error) {
+	data, err := ioutil.ReadFile("/tmp/netnscfg.log")
+	if err != nil {
+		return -1, err
+	}
+
+	subStr := "from PID "
+	netNsStr := string(data)
+	newStr := netNsStr[strings.Index(netNsStr, subStr)+len(subStr):]
+
+	var pidStr string
+	for _, char := range newStr {
+		if !unicode.IsDigit(char) {
+			break
+		}
+		pidStr += string(char)
+	}
+
+	pid, err := strconv.Atoi(pidStr)
+	if err != nil {
+		return -1, err
+	}
+	return pid, nil
+}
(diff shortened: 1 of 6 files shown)


dcantah:
Gonna leave this as a draft for now until I can verify a couple things

katiewasnothere:
@dcantah what's the status here?

dcantah:
@katiewasnothere This is ready for review (sort of). Need to change the logic for finding a network namespace to join now that we're no longer going to write out the /tmp/netnscfg.log file, but the rest is good to go.

internal/storage/cifs/cifs.go
```diff
@@ -0,0 +1,102 @@
...
+	}()
+
+	cmd := exec.Command(
+		"mount",
+		"-t",
+		"cifs",
+		source,
+		target,
+		"-o",
+		fmt.Sprintf("vers=3.0,username=%s,password=%s,%s", username, password, cifsMountOptions),
```
katiewasnothere, line 60: Why can't we use the storage `Mount` function or otherwise call unix.Mount here?

dcantah: We likely can, there was some reason the teleport folks shell out to the mount binary but I need to follow up with them again for this.

katiewasnothere: Ah I see, we issue this cmd to the network ns.

katiewasnothere: So with the new changes you made for `DoInNetNS`, we can construct a function here that makes use of the storage `Mount` call or unix.Mount.

dcantah: That shouldn't matter for this, the whole operation either via direct syscall or invoking this should all happen in the same netns. The teleport folks don't have this restriction and they invoke the binary, I just need to follow-up again for why.

dcantah: > So with the new changes you made for `DoInNetNS`, we can construct a function here that makes use of the storage `Mount` call or unix.Mount.

Yep

internal/storage/cifs/cifs.go
```diff
@@ -0,0 +1,102 @@
...
+	ns, err := netns.GetFromPid(pid)
+	if err != nil {
+		return errors.Wrapf(err, "netns.GetFromPid(%d) failed", pid)
+	}
+	defer ns.Close()
+
+	return network.DoInNetNS(ns, cmd.Run)
+}
+
+// This is gross, there HAS to be a better way
```
katiewasnothere, line 78: I take it this will be updated

dcantah: Yep!

internal/storage/loopback/loopback.go
```diff
@@ -0,0 +1,191 @@
...
+			osRemoveAll(target)
+		}
+	}()
+
+	// If the device already exists `getLoopID` will return -1, so skip this if the device
+	// we're asking for already exists. The UVM starts with 8 loop devices (0,1,2....7)
+	// already so this should be hit exactly 8 times before we have to actually allocate
+	// a new loop device.
+	loopID := device
+	if _, err := os.Stat(fmt.Sprintf("/dev/loop%d", device)); err != nil {
```
katiewasnothere, line 61: should we be checking that the error corresponds to no file being present here and then returning the error if not?

internal/storage/loopback/loopback.go
```diff
@@ -0,0 +1,191 @@
...
+		loopCtrl, err := getLoopController(context.Background())
```
katiewasnothere, line 62: shouldn't this be using the passed in `ctx` parameter?

internal/storage/loopback/loopback.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	log.G(ctx).WithField("deviceNumber", device).Debug("Removed loop device")
+	return int(device)
+}
+
+// PairLoopDevice pairs a backing file `backingFile` to the loop device with device number
+// `deviceNumber`. Returns the loop device path, e.g. "/dev/loop9"
+func pairLoopDevice(ctx context.Context, deviceNumber int, backingFile string) (loopName string, err error) {
+	loopName = fmt.Sprintf("/dev/loop%d", deviceNumber)
```
katiewasnothere, line 159: nit: consider passing in the loopName instead of the deviceNumber since we don't need it for anything else and the calling function already constructs the loopName.

internal/storage/loopback/loopback.go
```diff
@@ -0,0 +1,191 @@
...
+	source, err := pairLoopDevice(ctx, loopID, backingFile)
+	if err != nil {
+		return errors.Wrapf(err, "failed to pair backing file %s to loop device with ID %d", backingFile, loopID)
+	}
+
+	return unixMount(source, target, "ext4", uintptr(unix.MS_RDONLY), "noload")
+}
+
+// Teardown a file backed loop device
+func Teardown(ctx context.Context, device int) error {
```
katiewasnothere, line 84: is this not used?

internal/storage/loopback/loopback.go
```diff
@@ -0,0 +1,191 @@
...
+ err = ioctl(cfd, LOOP_CTL_REMOVE, devnr);
+
+ # find or allocate a free loop device to use
+ devnr = ioctl(cfd, LOOP_CTL_GET_FREE);
+
+ sprintf(loopname, "/dev/loop%i", devnr);
+ ffd = open("backing-file", O_RDWR);
+ lfd = open(loopname, O_RDWR);
+ err = ioctl(lfd, LOOP_SET_FD, ffd);
+*/
```
katiewasnothere, line 135: why?

internal/storage/loopback/safe.go
```diff
@@ -0,0 +1,120 @@
...
+	"golang.org/x/sys/unix"
+)
+
+type (
+	open struct {
+		path string
+		mode int
+		perm uint32
+	}
+
```
katiewasnothere, line 18: I've never seen this code structure before...

dcantah: This was copied as is from the daemon, I'm not entirely sure we even need the safe ioctl stuff. But me neither before this haha

katiewasnothere: what daemon?

dcantah: teleportd.

internal/storage/loopback/loopback.go
dcantah: This implements some functionality similar to what losetup would do under the hood. Another approach would be to mknod the device nodes but we still need the ioctls to add/remove the backing files

dcantah: The hope is we don't ever actually have to remove the device nodes and simply just remove the backing file and re-use the same node for another container

PR #393, opened 2021-02-19
Remove netnscfg gcstools utility

dcantah:
This utility only existed because go used to not play well with network namespaces
due to it's nature of goroutine thread multiplexing. This poor behavior was remedied/solved in go
1.10 due to some changes in how runtime.LockOSThread works with regards to scheduling new goroutines
and spawning new threads.

Great article on the problem: https://www.weave.works/blog/linux-namespaces-and-go-don-t-mix

and the eventual resolution: https://www.weave.works/blog/linux-namespaces-golang-followup

* Add DoInNetNS function to perform a function in a specific network namespace.
* Convert netnscfg to a function to be used in tandem with DoInNetNS


internal/network/netns.go
```diff
@@ -21,94 +10,100 @@ import (
...
-		log.Infof("Configure %s in %d with DHCP", *ifStr, *nspid)
+	// Move the interface to the new network namespace
+	if err := netlink.LinkSetNsPid(link, pid); err != nil {
+		return errors.Wrapf(err, "netlink.SetNsPid(%#v, %d) failed", link, pid)
 	}
+	return nil
+}
 
-	// Lock the OS Thread so we don't accidentally switch namespaces
+// DoInNetNS is a utility to run a function `work` inside of a specific network namespace
```
katiewasnothere, line 38: `run` instead of `work` here

internal/network/netns.go
```diff
@@ -21,94 +10,100 @@ import (
...
+}
+
+// NetNSConfig moves a network interface into a network namespace and
+// configures it. The configuration specified by `adapterCfg` is expected to be
+// passed in as a JSON object (marshalled prot.NetworkAdapter). This function will log
+// it's output to `logFile`.
+//
+// This function MUST be used in tandem with `DoInNetNS` or some other means that ensures that the goroutine
+// executing this code stays on the same thread.
+func NetNSConfig(ifStr string, nsPid int, adapterCfg string, logFile string) error {
```
katiewasnothere, line 68: Why not just pass in the prot.NetworkAdapter object here? The calling function marshals the object to a string and then this function just unmarshalls back

internal/network/netns.go
```diff
@@ -21,94 +10,100 @@ import (
...
+	if err != nil {
+		return errors.Wrap(err, "failed to create log file")
 	}
+	defer logFileHandle.Close()
 
-	// Move the interface to the new network namespace
-	if err := netlink.LinkSetNsPid(link, *nspid); err != nil {
-		return fmt.Errorf("netlink.SetNsPid(%#v, %d) failed: %v", link, *nspid, err)
+	log := logrus.New()
+	log.SetOutput(logFileHandle)
```
katiewasnothere, line 80: Do we need this now that we don't have to run a separate executable in the UVM? There's nothing that reads this file afaik anyways, I assume it's original intention was for debugging on our side.

dcantah: I wanted to keep it for debuggability but if no one cares or has ever found this useful I can just log as normal

dcantah: It's helped me in a few cases (but that's because I was specifically making changes TO this codepath 😆)

internal/network/netns.go
dcantah: Yea this is dumb, good call. Too much c+p

dcantah:
@katiewasnothere Remedied things, ptal

katiewasnothere (approved):
LGTM

PR #394, opened 2021-02-25
Ignore error when container process has exited between queries in /proc

katiewasnothere:
Recently we've been seeing failures when opengcs waits on a container's processes that says something like `failed to read cmdline file for process 195: open /proc/195/cmdline: no such file or directory.` As a result, opengcs believes the container has exited when it may not have. Eventually all threads waiting on the container are notified and continue shutdown logic for the container. 

This PR fixes this issue by ignoring errors if a process's cmdline information under /proc no longer exists between the time of first querying for all running processes and now. This PR also add more logging and verbose errors around querying for container processes to aid in further debugging. 

### DIFF
--- a/service/gcs/runtime/runc/runc.go
+++ b/service/gcs/runtime/runc/runc.go
@@ -21,6 +21,7 @@ import (
 	oci "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
 )
 
 const (
@@ -291,6 +292,11 @@ func (c *container) GetRunningProcesses() ([]runtime.ContainerProcessState, erro
 	for _, pid := range pids {
 		command, err := c.r.getProcessCommand(pid)
 		if err != nil {
+			if errors.Is(err, unix.ENOENT) {
+				// process has exited between getting the running pids above
+				// and now, ignore error
+				continue
+			}
 			return nil, err
 		}
 		pidMap[pid] = &runtime.ContainerProcessState{Pid: pid, Command: command, CreatedByRuntime: false, IsZombie: false}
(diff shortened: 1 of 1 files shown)


service/gcs/runtime/runc/runc.go
```diff
@@ -324,13 +325,20 @@ func (c *container) GetAllProcesses() ([]runtime.ContainerProcessState, error) {
 	if err != nil {
 		return nil, err
 	}
+	l := logrus.WithField("cid", c.id)
```
dcantah, line 328: Nit: Why not just `WithFields`? Might be missing something

dcantah: Realized I only selected one line, I'm referring to how you just end up adding another field right below and logging

service/gcs/runtime/runc/runc.go
```diff
@@ -324,13 +325,20 @@ func (c *container) GetAllProcesses() ([]runtime.ContainerProcessState, error) {
...
 	// and IsZombie at the default value of false.
 	for _, pid := range runningPids {
 		command, err := c.r.getProcessCommand(pid)
 		if err != nil {
+			if errors.Is(err, unix.ENOENT) {
+				// process has exited between getting the running pids above
+				// and now, ignore error
+				continue
+			}
 			return nil, err
```
dcantah, line 342: We don't want the same wrapped error here? https://github.com/microsoft/opengcs/pull/394/files#diff-e88ce53c5c78c98d276a0acc3ab891fb7c59db59a74f41262fa2ef160a84b4aaR373

service/gcs/runtime/runc/runc.go
```diff
@@ -324,13 +325,20 @@ func (c *container) GetAllProcesses() ([]runtime.ContainerProcessState, error) {
...
+	l := logrus.WithField("cid", c.id)
+	l.WithField("pids", runningPids).Debug("running container pids")
 
 	pidMap := map[int]*runtime.ContainerProcessState{}
 	// Initialize all processes with a pid and command, leaving CreatedByRuntime
 	// and IsZombie at the default value of false.
 	for _, pid := range runningPids {
 		command, err := c.r.getProcessCommand(pid)
 		if err != nil {
+			if errors.Is(err, unix.ENOENT) {
```
ambarve, line 337: I think we should add the same check in the `GetRunningProcesses` function that is just above this function?

kevpar:
What's the reasoning behind ignoring the `ENOENT` error? Do we know what causes it to occur?

katiewasnothere:
> What's the reasoning behind ignoring the `ENOENT` error? Do we know what causes it to occur?

Basically we query for all the running processes first thing entering `GetAllProcesses` and between doing that and later reading the pid's cmdline, a process may have died and been reaped, so that file would no longer exist. 

`ENOENT` is the error returned when trying to read a cmdline file for a process that no longer exists.  

Another example of this issue for a different project [here](https://gitlab.com/psmisc/psmisc/-/merge_requests/21).

service/gcs/runtime/runc/runc.go
katiewasnothere: I think I'm going to remove both wraps. We don't necessarily know at that point if the process *still* exists.

kevpar (approved):
LGTM

PR #398, opened 2021-03-26
Support passing through arbitrary options and propagation flags for scsi mounts

katiewasnothere:
This PR fixes the mount propagation behavior for scsi mounts. The expected behavior is that a `rshared` mount propagates mountpoint events from container to host and vice versa. However, since we previously did not create the initial mountpoint for the device in the UVM with the correct propagation flags, events would not flow from the container back to the UVM as expected. This PR fixes this by supporting the ability to pass in arbitrary mount options when attaching a scsi disk. 

### DIFF
--- a/internal/storage/mount.go
+++ b/internal/storage/mount.go
@@ -26,7 +26,74 @@ var (
 	unixMount   = unix.Mount
 	osRemoveAll = os.RemoveAll
 	listMounts  = listMountPointsUnderPath
+
+	flags = map[string]struct {
+		clear bool
+		flag  uintptr
+	}{
+		"acl":           {false, unix.MS_POSIXACL},
+		"async":         {true, unix.MS_SYNCHRONOUS},
+		"atime":         {true, unix.MS_NOATIME},
+		"bind":          {false, unix.MS_BIND},
+		"defaults":      {false, 0},
+		"dev":           {true, unix.MS_NODEV},
+		"diratime":      {true, unix.MS_NODIRATIME},
+		"dirsync":       {false, unix.MS_DIRSYNC},
+		"exec":          {true, unix.MS_NOEXEC},
+		"iversion":      {false, unix.MS_I_VERSION},
+		"lazytime":      {false, unix.MS_LAZYTIME},
+		"loud":          {true, unix.MS_SILENT},
+		"mand":          {false, unix.MS_MANDLOCK},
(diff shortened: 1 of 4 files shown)


internal/storage/scsi/scsi.go
```diff
@@ -78,6 +84,15 @@ func Mount(ctx context.Context, controller, lun uint8, target string, readonly b
 		}
 		break
 	}
+
+	if len(pgFlags) != 0 {
```
dcantah, line 88: Maybe comment above here explaining the need for the remount

dcantah (approved):
LGTM

internal/storage/scsi/scsi.go
```diff
@@ -53,15 +54,16 @@ func Mount(ctx context.Context, controller, lun uint8, target string, readonly b
...
 	var flags uintptr
 	data := ""
 	if readonly {
 		flags |= unix.MS_RDONLY
 		data = "noload"
 	}
-
 	for {
-		if err := unixMount(source, target, "ext4", flags, data); err != nil {
+		if err := unixMount(source, target, "ext4", uintptr(flags), data); err != nil {
```
dcantah, line 66: Why do we have to recast to uintptr? Isn't flags already the right type?

internal/storage/mount.go
```diff
@@ -26,8 +26,75 @@ var (
...
+		"remount":       {false, unix.MS_REMOUNT},
+		"ro":            {false, unix.MS_RDONLY},
+		"rw":            {true, unix.MS_RDONLY},
+		"silent":        {false, unix.MS_SILENT},
+		"strictatime":   {false, unix.MS_STRICTATIME},
+		"suid":          {true, unix.MS_NOSUID},
+		"sync":          {false, unix.MS_SYNCHRONOUS},
+	}
+
+	propagationFlags = map[string]int{
```
dcantah, line 69: nit: Maybe just have this be `map[string]uintptr` so we can avoid the cast for every remount here (same with flags on line 30 but for the `flag` field in the struct) https://github.com/microsoft/opengcs/pull/398/files#diff-92e6cc0c16687650cf45c183888d3bcb57ab32b12118044a82d305d11f6df12bR87

AFAIK none of these are negative values

dcantah (dismissed):
re-lgtm. Comments just on type casts that could be skipped

dcantah (approved):

PR #399, opened 2021-03-31
add support for dm verity target when mounting VPMem devices

anmaxvl:
Update VPMem mount APIs to support passing dm verity information
in addition to VPMem multi-mapping.

Hash device is expected to be the same as the data device with
hash tree appended right after the ext4 file system data, block
sizes for data and hash devices are expected to be the same as
well.

Additionally handle a case when both multi-mapping and verity
are enabled, in that case, first create dm-linear target and use
that target as a data and hash device for dm-verity target.

Signed-off-by: Maksim An <maksiman@microsoft.com>

### DIFF
--- a/internal/storage/pmem/pmem.go
+++ b/internal/storage/pmem/pmem.go
@@ -5,6 +5,8 @@ package pmem
 import (
 	"context"
 	"fmt"
+	"github.com/Microsoft/opengcs/internal/log"
+	"github.com/Microsoft/opengcs/service/gcs/prot"
 	"os"
 
 	"github.com/Microsoft/opengcs/internal/oc"
@@ -25,15 +27,19 @@ var (
 const (
 	pMemFmt         = "/dev/pmem%d"
 	linearDeviceFmt = "dm-linear-pmem%d-%d-%d"
+	verityDeviceFmt = "dm-verity-pmem%d-%s"
 )
 
-func mountInternal(source, target string) (err error) {
+// mountInternal mounts source to target via unix.Mount
+func mountInternal(ctx context.Context, source, target string) (err error) {
 	if err := osMkdirAll(target, 0700); err != nil {
 		return err
 	}
 	defer func() {
 		if err != nil {
-			osRemoveAll(target)
+			if err := osRemoveAll(target); err != nil {
+				log.G(ctx).WithError(err).Debugf("error cleaning up target: %s", target)
+			}
 		}
 	}()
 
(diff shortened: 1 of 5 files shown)


internal/storage/pmem/pmem.go
```diff
@@ -44,69 +50,168 @@ func mountInternal(source, target string) (err error) {
 	return nil
 }
 
-// Mount mounts the pmem device at `/dev/pmem<device>` to `target`.
+// Mount mounts the pmem device at `/dev/pmem<device>` to `target` in a basic scenario.
+// In the case when mappingInfo or/and verityInfo is/are not nil, device-mapper framework
+// is used to create linear and verity targets accordingly.
```
msscotb, line 55: If either mappingInfo or verifyInfo are non-nil, the device-mapper framework is used to create linear and verify targets accordingly. If both are non-nil, the linear target is created first and used as the data/hash device for the verity target.

msscotb: I'm not sure where this code explicitly causes the linear target to be used as the hash device for the verity target. Did I miss it or is this relationship actually established by the caller of Mount?

anmaxvl: it's [this] (https://github.com/microsoft/opengcs/blob/c1027f7c2c8b7b30dc38e204deaf9df773258a3a/internal/storage/pmem/pmem.go#L81-L107) portion of code that does it. Note that I'm reassigning `devicePath` variable, rather than creating a local one. now that I think about it, I should probably do the reassignment explicitly, since the code will be more clear.

internal/storage/pmem/pmem.go
```diff
@@ -44,69 +50,168 @@ func mountInternal(source, target string) (err error) {
...
-		trace.Int64Attribute("device", int64(device)),
+		trace.Int64Attribute("deviceNumber", int64(device)),
 		trace.StringAttribute("target", target))
 
-	source := fmt.Sprintf(pMemFmt, device)
-	return mountInternal(source, target)
+	devicePath := fmt.Sprintf(pMemFmt, device)
+	if mappingInfo == nil && verityInfo == nil {
+		return mountInternal(devicePath, target)
+	}
```
msscotb, line 77: This is redundant. The function has the same behavior with or without this block.

msscotb, line 77: This is redundant. The function has the same behavior with or without this block.

internal/storage/pmem/pmem.go
msscotb: Oh I missed that they were related by passing in the devicePath. Sure explicit reassignment might make it clearer though it probably isn't necessary.

internal/storage/pmem/pmem.go
```diff
@@ -44,69 +50,168 @@ func mountInternal(source, target string) (err error) {
...
+
+	return mountInternal(devicePath, target)
+}
+
+// createDMLinearTarget creates dm-linear target from a given `device` slot location and `mappingInfo`
+func createDMLinearTarget(ctx context.Context, devPath, devName string, target string, mappingInfo *prot.DeviceMappingInfo) (_ string, err error) {
+	if mappingInfo == nil {
+		return "", errors.New("mappingInfo cannot be nil")
+	}
```
katiewasnothere, line 116: This can never happen right?

internal/storage/pmem/pmem.go
```diff
@@ -44,69 +50,168 @@ func mountInternal(source, target string) (err error) {
...
+// 0 417792 verity 1 /dev/sdb /dev/sdc 4096 4096 52224 1 sha256 2aa4f7b7b6...f4952060e8 762307f4bc8...d2a6b7595d8..
+// |    |     |    |     |     |        |    |    |    |    |              |                        |
+// start|     |    |  data_dev |  data_block | #blocks | hash_alg      root_digest                salt
+//     size   |  version    hash_dev         |     hash_offset
+//          target                       hash_block
+func createDMVerityTarget(ctx context.Context, devPath, devName, target string, verityInfo *prot.DeviceVerityInfo) (_ string, err error) {
+	// regular mount if no verityInfo is present
+	if verityInfo == nil {
+		return "", errors.New("verityInfo cannot be nil")
+	}
```
katiewasnothere, line 151: same here. This shouldn't happen.

internal/storage/pmem/pmem.go
dcantah: You're right I believe (and for the verity info also). I always think about how we'd like to handle things like this because I feel like we run into this a lot. It's impossible we'd enter here currently but if this was ever reworked/expanded on so another time this is called wasn't guarded by mappingInfo being != nil in the first place then we're in a weird spot. The check is harmless though even if a bit redundant

internal/storage/pmem/pmem.go
```diff
@@ -25,15 +27,19 @@ var (
...
+// mountInternal mounts source to target via unix.Mount
 func mountInternal(source, target string) (err error) {
 	if err := osMkdirAll(target, 0700); err != nil {
 		return err
 	}
 	defer func() {
 		if err != nil {
-			osRemoveAll(target)
+			if err := osRemoveAll(target); err != nil {
+				log.G(context.TODO()).WithError(err).Debugf("error cleaning up target: %s", target)
```
dcantah, line 41: We could add a `ctx context.Context` parameter to `mountInternal` so we don't have to use a fresh context and just pass in the one we get from `Mount`

internal/storage/pmem/pmem.go
msscotb: My opinion is that it appropriate to just AV in cases that should never happen. One argument for this is it removes all confusion on whether the null is actually a possibility and needs to be handled.

internal/storage/pmem/pmem.go
msscotb: Hopefully it was clear that I'm referring to the if block only. The devicePath assignment is still needed.

internal/storage/pmem/pmem.go
katiewasnothere: What does AV stand for?

dcantah: Access violation I believe

kevpar: AKA segfault/nil point dereference

msscotb: Ah, I meant an Access Violation. In other words, go ahead and let it crash/panic/throw when it tries to deref the null we never expect vs checking for null and gracefully returning an error.

anmaxvl: ah, gocha! makes sense! I was pinging offline in parallel ahah

dcantah: Sounds good to me. At least from a containerd perspective and not a general best practices point of view, if we panic the cause is a little harder to debug than if we just bubbled up an error. Amit alleviated this by having panics actually get logged now though

msscotb (approved):

dcantah (approved):

